PR #14, opened 2026-02-06
Update README with Skills.sh installation guide

fellyph:
Added installation instructions for Skills.sh to the README and moved how it works before the installation process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,16 @@ To install multiple skills at once:
 npx skills add WordPress/agent-skills --skill wp-plugin-development wp-abilities-api wp-playground
 ```
 
+#### Choosing a scope
+
+`npx skills add` asks to choose the skill **project-scoped**, selecting the local scope on which the skills are installed in the local project; it can be stored at the repository (e.g. `.claude/skills/`, `.cursor/skills/`) — so the skills can be committed to version control and shared with your team.
+
+Installing **globally** makes the skill available to your user (across **all** your projects). Adding the `-g` / `--global` flag, it will install your skill with global scope:
+
+```bash
+npx skills add WordPress/agent-skills --skill wp-plugin-development --global
+```
+
 ### Install globally for Claude Code
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ You can use both at the same time. When a skill exists in both scopes, the proje
 
 ## Quick Start
 
-The fastest way to install skills is with a single command:
+The fastest way to install a skill is with a single command:
 
 ```bash
 npx skills add WordPress/agent-skills --skill wp-plugin-development
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ AI coding assistants are powerful, but they often:
 - Skip proper block deprecations, causing "Invalid block" errors
 - Ignore existing tooling in your repo
 
-Agent Skills solve this by giving AI assistants **expert-level WordPress knowledge** in a format they can actually use.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,8 +34,44 @@ Agent Skills solve this by giving AI assistants **expert-level WordPress knowled
...
+
+When you ask your AI assistant to work on WordPress code, it reads these skills and follows the documented procedures rather than guessing.
+
+
 ## Quick Start
 
+The fastest way to install skills is with a single command:
+
+```bash
+npx skills add WordPress/agent-skills --skill wp-plugin-development
```
brandonpayton, line 74: @fellyph this seems like a good change, but let's add clarification about installing global and project scopes.

README.md
```diff
@@ -14,7 +14,7 @@ AI coding assistants are powerful, but they often:
 - Skip proper block deprecations, causing "Invalid block" errors
 - Ignore existing tooling in your repo
 
-Agent Skills solve this by giving AI assistants **expert-level WordPress knowledge** in a format they can actually use.
+Agent Skills solves this by giving AI assistants **expert-level WordPress knowledge** in a format they can actually use.
```
brandonpayton: I think the previous text was more grammatically correct. It's a plural collection of skills.

README.md
```diff
@@ -34,8 +34,58 @@ Agent Skills solve this by giving AI assistants **expert-level WordPress knowled
...
+To see all available skills:
+
+```bash
+npx skills add WordPress/agent-skills --list
+```
+
+To install multiple skills at once:
+
+```bash
+npx skills add WordPress/agent-skills --skill wp-plugin-development wp-abilities-api wp-playground
```
brandonpayton, line 86: @fellyph we are also recommending installation via our own custom `shared/scripts/skillpack-install.mjs` script below in addition to recommending `npx skills` here.

I'd personally prefer using a general purpose command for installation that is maintained by the community since it is common logic for those installing skill sets. But either way, it seems like it would be good to pick one.

Is there something the custom scripts do that `npx skills` doesn't do? If not, I think we should check with the AI team about whether they'd be up for switching to `npx skills`.

I think the custom script has been there from the beginning when @Jameswlepage hard-reset this repo and contributed a large set of WP agent skills (from under the Automattic GH org).

fellyph: I think we get more people installing via npx skills, a better position on popular skills, and more visibility.

jonathanbossenger: @mikeyarce, what do you think here? I'm personally in favor of recommending the `npx skills add` path (via https://www.skills.sh/) as it makes it really easy for newcomers to add skills.

mikeyarce: Yeah I think we should also surface `npx skills add` as an option

README.md
mikeyarce: Yeah agree with global / local scopes clarification.  Should the main example also have a specific skill when the instruction says `The fastest way to install skills` but we're only installing one skill?

mikeyarce (commented):
Added a few comments - I like the idea of `npx skills add`

README.md
jonathanbossenger: @fellyph just checking if this change has been made or not? If it has, please mark this as resolved.

README.md
jonathanbossenger: @fellyph can this be marked as resolved?

jonathanbossenger (approved):
LGTM, just waiting to resolve the two review comments.

mikeyarce (approved):
Looks good to me!

PR #15, opened 2026-02-06
docs(wp-interactivity-api): Add WordPress 6.9 client navigation requirements

derintolu:
## Summary

Adds documentation for the WordPress 6.9 requirement that script modules must be explicitly marked as compatible with client-side navigation.

Per [Interactivity API's client navigation improvements in WordPress 6.9](https://make.wordpress.org/core/2025/11/12/interactivity-apis-client-navigation-improvements-in-wordpress-6-9/):

> For other script modules registered "manually", developers must use the `add_client_navigation_support_to_script_module` method of the main `WP_Interactivity_API` instance.

Without this, the router will not intercept link clicks on pages using manually registered script modules.

## Changes

- **SKILL.md**: Updated client-side navigation section with WP 6.9+ requirements
- **references/router.md**: Added new reference document with full router documentation

## Key Addition

```php
// Mark module as compatible with client-side navigation (WordPress 6.9+ REQUIRED)
wp_interactivity()->add_client_navigation_support_to_script_module('my-navigation');
```

For blocks, use `block.json`:
```json
{
  "supports": {
    "interactivity": { "clientNavigation": true }
  }
}
```

## Test Plan

- [x] Verified this requirement on a production WordPress 6.9 site
- [x] Confirmed router does not intercept links without this call
- [x] Confirmed router works correctly after adding this call

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/skills/wp-interactivity-api/SKILL.md
+++ b/skills/wp-interactivity-api/SKILL.md
@@ -13,7 +13,8 @@ Use this skill when the user mentions:
 - Interactivity API, `@wordpress/interactivity`,
 - `data-wp-interactive`, `data-wp-on--*`, `data-wp-bind--*`, `data-wp-context`,
 - block `viewScriptModule` / module-based view scripts,
-- hydration issues or “directives don’t fire”.
+- hydration issues or "directives don't fire",
+- **client-side navigation**, `@wordpress/interactivity-router`, `data-wp-router-region`.
 
 ## Inputs required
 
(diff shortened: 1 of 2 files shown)


skills/wp-interactivity-api/SKILL.md
```diff
@@ -134,16 +135,55 @@ Verify the repo supports the required module build path:
 - if it uses `@wordpress/scripts`, prefer its conventions.
 - if it uses custom bundling, confirm module output is supported.
 
-### 6) Debug common failure modes
+### 6) Client-side navigation with the Router
```
justlevine, line 138: TBF this could be asked about _all of the existing committed skills_, but what's the point of dumping all this information here if it's available via references?

skills/wp-interactivity-api/SKILL.md
```diff
@@ -174,6 +214,7 @@ See `references/debugging.md`.
 
 - If repo build constraints are unclear, ask: "Is this using `@wordpress/scripts` or a custom bundler (webpack/vite)?"
 - Consult:
+  - `references/router.md` - **READ THIS FIRST for navigation issues**
```
justlevine, line 217: We should avoid yelling out our models since this [needs to service more than just Anthropic](https://wordpress.slack.com/archives/C08TJ8BPULS/p1770395934195279)

jonathanbossenger:
@derintolu, thank you for your pull request to improve WordPress agent-skills.

I used AI to review this PR, and it suggested these changes. Let me know what you think.

- This PR adds a new procedure step (step 6) for client-side navigation with the router, new debugging steps for navigation issues, and new trigger phrases in "When to use". This is a behavioral change — an agent receiving a query like "my router navigation stopped working after upgrading to WP 6.9" would now produce different output. The existing eval scenario (interactivity-route-and-debug.json) covers directive debugging but not navigation-specific queries. Consider adding or updating a scenario to cover the router/navigation path.
- Step 6 in SKILL.md contains substantial inline content about the router (4-step setup, how automatic interception works, the WP 6.9 requirement, common mistakes) and then says "See references/router.md for full details." The reference file repeats much of the same content in more detail. Step 6 could be trimmed to the essential procedure (enqueue, mark compatible, add region, use plain links) with a clear pointer to references/router.md for the why/how/debugging. This would reduce duplication and keep SKILL.md closer to a checklist.
- The compatibility field now says "Targets WordPress 6.9+" but the skill's existing content covers features from WordPress 6.5+. The 6.9 requirement only applies to the new client-navigation compatibility marking. Consider clarifying this as "Targets WordPress 6.5+ (client-side navigation features require 6.9+)" to avoid implying the entire skill requires 6.9.

Other nitpick items:
- The reference file references/router.md uses emoji markers for code examples. Existing reference files in the repo tend to use plain text headers. Minor style consistency point.
- The "CRITICAL" and "REQUIRED" emphasis in step 6 is quite strong. The upstream WordPress docs present this as a standard requirement, not an emergency. Consider toning down to match the register of the rest of the skill.

jonathanbossenger:
@derintolu there has been no movement on this PR for two weeks, so I'm closing it out. Feel free to reopen it if you'd like to move it forward.

PR #24, opened 2026-02-25
Block development: static vs dynamic blocks

luisherranz:
## What

Teach the block-development skill how to choose between static and dynamic rendering when creating a new block, and add eval scenarios to verify the decision.

## Why

The skill had no guidance on this decision. Agents would default to whichever pattern they'd seen most in training data, leading to inconsistent choices: sometimes static when dynamic was clearly needed (e.g., Interactivity API blocks, blocks in template parts), and sometimes dynamic for simple presentational blocks where static is fine. A priority-ordered checklist eliminates the guesswork.

## How

- Add `references/static-vs-dynamic-blocks.md`: a "first YES wins" decision checklist covering external data, evolving markup, Interactivity API, block themes, server-side needs, content portability, and simple presentational blocks.
- Update `SKILL.md` step 3 ("Pick the right block model") to point to the new reference and simplify the interactivity guidance into a single line that loads `wp-interactivity-api`.
- Add 8 new eval scenarios (6 dynamic, 2 static) that each test a different decision point from the checklist, including one adversarial case where a lower-priority signal (portability) conflicts with a higher-priority one (server-side auth).
- Update the existing `block-create-interactive-inner-blocks` eval to include the static-vs-dynamic decision step, since Interactivity API requires dynamic rendering.


luisherranz:
@Jameswlepage, any guidance on how to run the evals to test that they work?

karmatosed:
@luisherranz right now this branch is out of date, so I would suggest working on that first then we can approach the evals. That is of course if you want to work on getting this in still, if not please comment and we can close. I am going through and trying to triage some cleanup.

jonathanbossenger:
FWIW the skill itself looks good, so we just need to update the branch.

luisherranz:
@copilot can you fix the CI problem?

skills/wp-block-development/references/static-vs-dynamic-blocks.md
```diff
@@ -0,0 +1,35 @@
...
+Ask these questions in order — the first YES wins:
+
+1. Does the block depend on data outside its own attributes (latest posts, user info, site options, database queries)?
+   → **Dynamic.**
+
+2. Will the block's markup likely change over time (design updates, bug fixes, accessibility improvements)?
+   → **Dynamic.** Static blocks freeze markup in the database; updating it later requires deprecations and manual re-saves.
+
+3. Will the block use the Interactivity API?
+   → **Dynamic.** Hydration requires server-rendered HTML to match the current code. Stale markup stored by a static block can cause mismatches.
```
mikeyarce, line 16: The Interactivity API supports both static and dynamic blocks, so this rule might not help.

skills/wp-block-development/references/static-vs-dynamic-blocks.md
```diff
@@ -0,0 +1,35 @@
...
+
+4. Is the block part of Block Themes (templates, template parts, theme blocks)?
+   → **Dynamic.** Block themes replace PHP themes, which are architecturally server-rendered.
```
mikeyarce, line 19: If a block is part of a block theme, why does it have to be dynamic?  I'm not totally sure I understand here.

eval/scenarios/block-choose-dynamic-for-team-members.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "Choose dynamic for block with external data",
+  "skills": ["wordpress-router", "wp-project-triage", "wp-block-development"],
+  "query": "Create a block that shows our team members with their photo, role and bio.",
+  "expected_behavior": [
```
mikeyarce, line 5: I'm not too sure about this eval - the prompt doesn't say where the "team members" comes from, and it could be just attributes inside of a static block, right?  Maybe we could clairfy the data source of team members so that we know it should be dynamic?

mikeyarce (changes requested):
Left a couple questions!

skills/wp-block-development/references/static-vs-dynamic-blocks.md
luisherranz: In theory it does, but in static blocks, any change in behavior breaks blocks that are already saved in the database. So using the Interactivity API with static blocks is not recommended and should only be reserved for very advanced cases. In those cases, the developer and agents can reach that conclusion on their own. For the skill, it is preferable that dynamic blocks are always used by default.

skills/wp-block-development/references/static-vs-dynamic-blocks.md
luisherranz: There is no benefit in making blocks that are not going to be saved in `post_content` static, as the main advantage of static blocks is that the user could one day export the content directly from the database and still keep something representative without the need to execute WordPress PHP. As that is not necessary for blocks whose purpose is not to be saved inside `post_content`, but rather belong to what in classic themes would have been rendered with PHP (and would never have been exported), there is no advantage in using static blocks, only all of their issues/disadvantages (updating a block means going to all the templates that use it and clicking save again, maintaining block deprecations, etc...).

eval/scenarios/block-choose-dynamic-for-team-members.json
luisherranz: Something like this? https://github.com/WordPress/agent-skills/pull/24/commits/898b5d9694af2bd2173f0e6d78b96b6d159308ed

PR #28, opened 2026-03-10
Fix outdated callback parameter name in PHP registration example

sudoleg:
The code example and parameter table in references/php-registration.md use callback, which is incorrect. The current WordPress Abilities API uses `execute_callback` as confirmed by the [official PHP reference](https://developer.wordpress.org/apis/abilities-api/php-reference/#parameters-explained). This caused the example to produce non-functional ability registrations.
                  
Changed in two places: the wp_register_ability() code snippet and the key arguments table.


galatanovidiu (approved):
Nice catch. Verified this against WP core (`wp-includes/abilities-api.php`, line 240) — the parameter is `execute_callback`, not `callback`. The current skill would produce broken registrations, so this is a good fix.

Clean and minimal, nothing else to flag. 👍

Issue #9, opened 2026-02-05
What other skill should we add?


Jameswlepage:
Interested in folks' ideas on what other skills would make sense to bring to this WordPress repository.

juanmaguitar:
Skills to help contributors with tasks related to WordPress DotOrg project. For example triaging issues.

ben104:
would be great if there were skills around checking the code, so for example when you build a plugin with wp-plugin-development you could then run a wp-plugin-check skill, which would work similarly to the official 'Plugin Check Plugin (PCP)' it could improve code quality, or is this be better in the plugin development skill?

Xilonz:
I cant think of skills specifically for WordPress core.

I'm not sure if this is in scope (yet):
How do you envision third-party plugin authors extending this? For example, if a plugin author wants to register a custom skill specific to their plugin (like "Clear Cache" or "Working with products"). 

In my opinion, it would be beneficial to this project to integrate into the WordPress ecosystem.

flexseth:
I'd love to see individual skills that create, per the documentation, WordPress Core Components.

One of the greatest discrepancies I've seen is the basic aspect of creating a component.

Between Storybook, Devhub, Gutenberg repo (has examples), and the documentation for Core Components, is a lack of consistency. Which makes sense because things change and very few people get paid for writing documentation  :D 

Having skills that build Core Components the WordPress Way would be 💯

gcapnias:
In my opinion, the ability to convert plain HTML into WordPress Block Markup is an essential skill.  
- When creating themes, part of the work involves transforming CSS-styled HTML into WordPress Block Markup for pattern creation.  
- Since there’s no official documentation on WordPress Block Markup, developers often need to search the block registry to find block names before they can use them.

juanmaguitar:
> Skills to help contributors with tasks related to WordPress DotOrg project. For example triaging issues.

Pinging @sirreal as he has developed some interesting WP-related AI tools:  https://github.com/sirreal/agent-skills

sirreal:
I've been working on a [WordPres Trac Claude plugin](https://github.com/sirreal/agent-skills/tree/main/plugins/wordpress-trac). I has skills for getting information from Trac tickets and changesets as well as for searching Trac or exploring the timeline.

[I'd love to get an MCP plugin on Trac](https://github.a8c.com/darthhexx/trac-mcp-plugin) that could make the whole plugin redundant.

owlot:
[...]

PR #38, opened 2026-04-09
Add wp-env skill for local WordPress development

rushikeshmore:
## What?

Adds a new `wp-env` skill that teaches AI coding assistants how to set up, configure, and troubleshoot local WordPress development environments using `@wordpress/env`.

Addresses the community request in #9 (comment from @owlot: "Would love to see an wp-env skill").

## Why?

`@wordpress/env` is the official Docker-based local development tool for WordPress plugin and theme development. It's referenced throughout the Gutenberg docs and used by core contributors, but there's no agent skill covering it. AI assistants currently guess at wp-env configuration and often make mistakes like using bare plugin slugs instead of ZIP URLs, or not checking Docker status before starting.

## How?

**New skill:** `skills/wp-env/SKILL.md`
- All 7 required sections (When to use, Inputs required, Procedure, Verification, Failure modes, Escalation)
- Covers: installation, auto-detection (plugin/theme/core), `.wp-env.json` configuration, source string formats, WP-CLI commands, PHPUnit testing, Xdebug setup, multisite, override files, environment management
- Common gotchas table (port conflicts, bare slugs, override merge behavior, Docker not running)

**Eval scenarios:** 2 JSON scenarios
- `wp-env-start-plugin.json` -- setting up wp-env for a plugin project
- `wp-env-troubleshoot-port.json` -- diagnosing and fixing port conflicts

**README:** Updated Available Skills table with wp-env entry.

## Testing

- [x] `node eval/harness/run.mjs` passes
- [x] `node shared/scripts/skillpack-build.mjs --clean` builds successfully for all 4 targets (codex, vscode, claude, cursor)
- [x] Skill follows required SKILL.md structure with all 7 sections
- [x] Two eval scenarios added
- [x] Frontmatter validates (name matches folder, compatibility includes WP 6.9 + PHP 7.2.24+)

## Use of AI Tools

Claude Code was used to assist with drafting this skill. Content was verified against the official [@wordpress/env documentation](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env/) and [source code](https://github.com/WordPress/gutenberg/tree/trunk/packages/env).

### DIFF
--- a/skills/wp-env/SKILL.md
+++ b/skills/wp-env/SKILL.md
@@ -24,6 +24,8 @@ Zero-config, Docker-based local WordPress development environment for plugins, t
 - this is a WordPress core or Gutenberg checkout
 - the user explicitly asks for wp-env or Docker
 
+wp-env can also run without Docker by using Playground as its runtime (`npx @wordpress/env start --runtime=playground`, experimental). That is still wp-env, driven by the same `.wp-env.json`, so treat it as a Docker-free fallback for a project that already carries wp-env config, not as a third option. It swaps MySQL for SQLite and drops `wp-env run` and the separate tests environment, which are most of the reasons to pick wp-env in the first place. For a quick Docker-free WordPress with no wp-env config, use the **wp-playground** skill directly.
+
 ## Inputs required
 
 1. **Docker status** -- verify Docker is installed and running: `docker info`
--- a/skills/wp-env/SKILL.md
+++ b/skills/wp-env/SKILL.md
@@ -1,0 +1,220 @@
+---
+name: wp-env
+description: "Use when setting up, configuring, or troubleshooting local WordPress development environments with @wordpress/env (wp-env). Triggers on mentions of wp-env, local WordPress development, Docker-based WordPress, or requests to start/stop/configure a local WordPress instance."
+compatibility: "Targets WordPress 7.0+ (PHP 7.4.0+). Requires Docker and Node.js 18.12+."
+---
+
+# @wordpress/env (wp-env)
+
+Zero-config, Docker-based local WordPress development environment for plugins, themes, and core.
+
+## When to use
+
+- User asks to set up a local WordPress development environment
+- Project contains a `.wp-env.json` file
+- User mentions `wp-env`, `@wordpress/env`, or Docker-based WordPress development
+- User wants to run WP-CLI commands, PHPUnit tests, or debug with Xdebug locally
+- Triage detects a plugin or theme project that needs a local WordPress instance
+
+**wp-env or wp-playground?** For a quick local WordPress, prefer the **wp-playground** skill by default — it's faster, disposable, and needs no Docker. Use **wp-env** when the task actually requires it:
+
+- the repo already contains a `.wp-env.json`
+- a real MySQL database is needed (PHPUnit against the DB, `wp db` commands, data that survives restarts)
+- arbitrary commands must run inside the environment (shell access, Composer, full WP-CLI via `wp-env run`)
+- this is a WordPress core or Gutenberg checkout
+- the user explicitly asks for wp-env or Docker
+
+## Inputs required
+
+1. **Docker status** -- verify Docker is installed and running: `docker info`
+2. **Node.js version** -- must be >= 18.12.0: `node -v`
+3. **Project type** -- plugin, theme, or full site (check for `.wp-env.json`, plugin headers, or `style.css` theme headers)
+4. **Existing config** -- read `.wp-env.json` and `.wp-env.override.json` if present
+
+## Procedure
+
+### 1. Install wp-env
+
+```sh
+# Global (recommended)
+npm -g install @wordpress/env
+
+# Or project-local
+npm i @wordpress/env --save-dev
+# Then use: npx wp-env start
+```
+
+### 2. Start the environment
+
+```sh
+wp-env start
+```
+
+Default credentials:
+- **URL:** http://localhost:8888/wp-admin/
+- **Username:** `admin`
+- **Password:** `password`
+
+Common start options:
+- `wp-env start --update` -- pull latest sources and reconfigure
+- `wp-env start --xdebug` -- enable Xdebug (debug mode)
+- `wp-env start --xdebug=profile,trace` -- multiple Xdebug modes
+- `wp-env start --auto-port` -- find available ports when defaults are busy
+
+### 3. Auto-detection (no config file)
+
+When no `.wp-env.json` exists, wp-env scans the current directory:
+
+| Detected type | How detected | Auto-config |
+|--------------|-------------|-------------|
+| Plugin | `Plugin Name:` header in a root `.php` file | `{ "plugins": ["."] }` |
+| Theme | `Theme Name:` header in `style.css` | `{ "themes": ["."] }` |
+| Core | `wp-includes/version.php` exists | `{ "core": "." }` |
+
+### 4. Configure with `.wp-env.json`
+
+Place at the project root. All fields are optional.
+
+```json
+{
+  "core": null,
+  "phpVersion": "8.1",
+  "plugins": [
+    ".",
+    "https://downloads.wordpress.org/plugin/akismet.zip",
+    "WordPress/classic-editor"
+  ],
+  "themes": [],
+  "port": 8888,
+  "multisite": false,
+  "phpmyadmin": false,
+  "config": {
+    "WP_DEBUG": true,
+    "SCRIPT_DEBUG": true
+  },
+  "mappings": {
+    "wp-content/mu-plugins": "./mu-plugins"
+  },
+  "lifecycleScripts": {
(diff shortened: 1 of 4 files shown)


rushikeshmore:
Hi, just checking if this is ready for review or if anything needs changing on my end. Happy to adjust.

skills/wp-env/SKILL.md
```diff
@@ -0,0 +1,212 @@
+---
+name: wp-env
+description: Use when setting up, configuring, or troubleshooting local WordPress development environments with @wordpress/env (wp-env). Triggers on mentions of wp-env, local WordPress development, Docker-based WordPress, or requests to start/stop/configure a local WordPress instance.
+compatibility: Targets WordPress 6.9+ (PHP 7.2.24+). Requires Docker and Node.js 18.12+.
```
mikeyarce:
```suggestion
compatibility: "Targets WordPress 7.0+ (PHP 7.4.0+). Requires Docker and Node.js 18.12+."
```

skills/wp-env/SKILL.md
```diff
@@ -0,0 +1,212 @@
+---
+name: wp-env
+description: Use when setting up, configuring, or troubleshooting local WordPress development environments with @wordpress/env (wp-env). Triggers on mentions of wp-env, local WordPress development, Docker-based WordPress, or requests to start/stop/configure a local WordPress instance.
```
mikeyarce:
```suggestion
description: "Use when setting up, configuring, or troubleshooting local WordPress development environments with @wordpress/env (wp-env). Triggers on mentions of wp-env, local WordPress development, Docker-based WordPress, or requests to start/stop/configure a local WordPress instance."
```

skills/wp-env/SKILL.md
```diff
@@ -0,0 +1,212 @@
...
+| "Cannot connect to Docker daemon" | Docker not running | Start Docker Desktop |
+| "Port 8888 already in use" | Port conflict | Use `--auto-port` or set custom `port` in `.wp-env.json` |
+| Plugin not appearing | Missing `Plugin Name:` header in main PHP file | Add standard plugin header comment |
+| "Could not find a valid source" | Invalid source string in config | Use full ZIP URL for wp.org plugins, not bare slugs |
+| Stale environment after source changes | Cached Docker volumes | `wp-env start --update` or `wp-env destroy && wp-env start` |
+| White screen / PHP errors | Corrupted database | `wp-env reset all && wp-env start` |
+| Override not taking effect | Wrong merge behavior | `plugins`/`themes` in override **replace** base arrays; only `config`/`mappings` merge |
+| Tests environment not accessible | Wrong port | Test environment runs on port 8889 by default |
+| Xdebug not connecting | IDE not listening or wrong port | Ensure IDE listens on port 9003 with correct `pathMappings` |
+| npm global install permission error | Node installed without nvm | Use `nvm` or prefix with `sudo` (not recommended) |
```
mikeyarce: I'd probably drop the `sudo` reccommendation here

```suggestion
| npm global install permission error | Node installed to a system path | Use `nvm`, or install locally: `npm i -D @wordpress/env` and run via `npx wp-env` |
```

skills/wp-env/SKILL.md
```diff
@@ -0,0 +1,212 @@
...
+
+Zero-config, Docker-based local WordPress development environment for plugins, themes, and core.
+
+## When to use
+
+- User asks to set up a local WordPress development environment
+- Project contains a `.wp-env.json` file
+- User mentions `wp-env`, `@wordpress/env`, or Docker-based WordPress development
+- User wants to run WP-CLI commands, PHPUnit tests, or debug with Xdebug locally
+- Triage detects a plugin or theme project that needs a local WordPress instance
```
mikeyarce, line 17:
```suggestion
- Triage detects a plugin or theme project that needs a local WordPress instance

**wp-env or wp-playground?** For a quick local WordPress, prefer the **wp-playground** skill by default — it's faster, disposable, and needs no Docker. Use **wp-env** when the task actually requires it:

- the repo already contains a `.wp-env.json`
- a real MySQL database is needed (PHPUnit against the DB, `wp db` commands, data that survives restarts)
- arbitrary commands must run inside the environment (shell access, Composer, full WP-CLI via `wp-env run`)
- this is a WordPress core or Gutenberg checkout
- the user explicitly asks for wp-env or Docker
```

mikeyarce (changes requested):
Thanks for the contribution @rushikeshmore !  

Added a few notes that I think we need to work on before this is ready.

I think one of my concerns is having multiple "dev envs" like wp-env and playground, so I suggested adding some wording to help the agent know which to use when.

Once this is merged I'll make a follow-up PR with some similar instructions on the wp-playground side.

skills/wp-env/SKILL.md
rushikeshmore: Applied. Worth noting this was more than style: the harness now requires "WordPress 7.0" and "PHP 7.4.0" in compatibility after #70, so the old 6.9 / 7.2.24 string fails validation on current trunk.

skills/wp-env/SKILL.md
rushikeshmore: Applied.

skills/wp-env/SKILL.md
rushikeshmore: Applied. Agreed, recommending sudo for an npm global install is bad advice to leave in a skill an agent will follow.

skills/wp-env/SKILL.md
rushikeshmore: Applied as written.

One knock-on effect: the plugin eval scenario asked "set up a local WordPress development environment for my plugin so I can test it", which under this new guidance should route to wp-playground, not wp-env. I updated that scenario so it does not contradict the skill: the query now asks for PHPUnit against a real database, and choosing wp-env over wp-playground is an explicit step plus a success criterion. Happy to revert that part if you would rather keep the scenarios untouched.

Agreed on the follow-up for the wp-playground side.

rushikeshmore:
Thanks for the review. All four suggestions applied, and the branch is rebased on trunk so the conflict is gone.

- compatibility bumped to WordPress 7.0+ / PHP 7.4.0+ (also required by the harness after #70)
- description quoted to match the other skills
- dropped the sudo recommendation, points at nvm or a local install instead
- added the wp-env vs wp-playground routing block

One extra change beyond the suggestions: the plugin eval scenario used a generic "set up a local environment" query that the new routing block would send to wp-playground. I made it a real wp-env case (PHPUnit against a database) and added the routing decision as a step. Details in the thread, easy to revert if you want the scenarios left alone.

`node eval/harness/run.mjs` passes.

mikeyarce (approved):
Thanks @rushikeshmore !

@bgrgicak can you take a look?  I know you worked on the Playground skills and would like to have your opinion here.  Specifically, do you think the routing will be good here?  I think we should only use wp-env when people ask directly or when it's needed for things like PHPUnit etc.

skills/wp-env/SKILL.md
```diff
@@ -0,0 +1,220 @@
...
+# @wordpress/env (wp-env)
+
+Zero-config, Docker-based local WordPress development environment for plugins, themes, and core.
+
+## When to use
+
+- User asks to set up a local WordPress development environment
+- Project contains a `.wp-env.json` file
+- User mentions `wp-env`, `@wordpress/env`, or Docker-based WordPress development
+- User wants to run WP-CLI commands, PHPUnit tests, or debug with Xdebug locally
```
bgrgicak, line 16: I'm not sure if it matters, Playground CLI can run WP-CLI, PHPUnit and Xdebug, so there aren't blockers.
I would leave this section as is.

bgrgicak:
@mikeyarce sorry for the late reply, I was afk for a while.

Preferring wp-env over Playground for development makes sense. 
The routing for when to choose wp-env seems good.

There's just one thing I would clarify. [WP-env supports Playground as a runtime](https://make.wordpress.org/playground/2026/02/06/wp-env-now-runs-wordpress-with-playground-runtime/). This could be confusing to the agent if we don't clarify what to use and when.

skills/wp-env/SKILL.md
rushikeshmore: Agreed, left this section as is.

rushikeshmore:
Thanks @bgrgicak, good catch. Added a short paragraph under the routing block (ad1082d) that names the Playground runtime explicitly: `--runtime=playground` is still wp-env driven by the same `.wp-env.json`, so it is a Docker-free fallback for projects that already carry wp-env config, not a third option. It also spells out what that runtime gives up (SQLite instead of MySQL, no `wp-env run`, no separate tests environment) based on the make/playground post, and points to the wp-playground skill for a quick Docker-free WordPress with no wp-env config.

Left the "When to use" bullet unchanged per your inline note. Rebased on trunk, harness passes.

bgrgicak (approved):
Thanks @rushikeshmore

PR #44, opened 2026-04-30
wp-abilities-api: add three strategy-layer references (domain-vs-projection, shared-core-service, grouping-heuristic)

rtio:
Adds three reference docs to the `wp-abilities-api` skill plus a six-line cross-link block in `SKILL.md`, and refreshes the existing `php-registration.md` reference against `class-wp-ability.php` trunk.

## What lands

| File | Purpose |
|---|---|
| `references/domain-vs-projection.md` | Three-layer model (domain capability / workflow / projection) and the use-case-contract test ("would a human intentionally do this through a supported UI or workflow?"). The projection-layer example anchors on the bundled `WordPress/mcp-adapter`'s `discover-abilities` / `execute-ability` / `get-ability-info` meta-abilities. |
| `references/shared-core-service.md` | When to delegate to a REST controller vs. extract a shared service. **Default: shared service.** Delegation is a conditional shortcut for low-stakes reads — three conditions enumerated. Six side-effect categories (telemetry, audit logs, custom hooks, notifications, cache writes, first-call REST bootstrap cost) that disqualify the shortcut. Plugin-family override section noting payments / subscriptions / own-MCP-transport plugins routinely tighten further. |
| `references/grouping-heuristic.md` | Granularity rules (semantic-intent vs REST-atomization vs action-bundle) with two generic worked examples. |
| `references/php-registration.md` | Refreshed: `permission_callback` is required (throws on missing); `meta.annotations` keys (`readonly`, `destructive`, `idempotent`) framed as Strongly Recommended; `meta.mcp.public` / `meta.mcp.type` taught for visibility through the bundled MCP adapter. |
| `wp-abilities-api/SKILL.md` | Six lines: layer-model intro at `## Procedure` + step-4 cross-links. |

## Verification at tip

- WP core symbols (`WP_REST_Request`, `WP_Error`, `rest_ensure_response`, `__()`, `rest_do_request`, `set_query_params`, `is_error`, `as_error`) verified against `WordPress/wordpress-develop` trunk.
- mcp-adapter abilities verified at `WordPress/mcp-adapter/includes/Abilities/{Discover,Execute,GetAbilityInfo}Ability.php`.

## Status

Round 4 (current tip, ~417 added lines). Iteration:

- **Round 2** — @nerrad's six inline comments addressed (registration-guide refresh, broadened inclusion test beyond admin-only, removed Jetpack Forms reference, prose-only delegate-helper framing, telemetry demoted to one of five side-effects).
- **Round 3** — three substantive deltas from pilot-ability learnings (`meta.mcp.public` / `mcp.type`; annotations escalated to Strongly Recommended; first-call REST bootstrap cost added as the sixth side-effect category).
- **Round 4** — @nerrad's two 2026-05-21 review items: mcp-adapter projection example anchored (`25530c6`); REST delegation reframed as a conditional shortcut, shared service named as default (`479c48c`).

### DIFF
--- a/skills/wp-abilities-api/references/domain-vs-projection.md
+++ b/skills/wp-abilities-api/references/domain-vs-projection.md
@@ -1,0 +1,111 @@
+# Domain capability vs projection — picking the layer to register at
+
+> Three layers in this model: **domain**, **projection**, and an optional **workflow** layer that composes abilities into multi-step tasks. The title names the two primary decisions; workflow is introduced where it earns its keep.
+
+The Abilities API can be used at two heights. You can register abilities as the surface an MCP/REST/Command-Palette client will consume directly. Or you can register abilities at the domain layer — "what can this plugin do, transport-neutrally?" — and let each consumer see a projection of that surface that suits its constraints. This reference covers why the second framing pays off, the three-layer model that operationalizes it, and how to use it when deciding what to register.
+
+## Why this matters
+
+Treating MCP exposure as the registration target conflates two decisions:
+
+1. What capabilities does this plugin offer?
+2. How should each consumer see them?
+
+The first decision is stable. The second is volatile — token budgets shift, MCP clients improve, Command Palette gains new conventions, large plugins outgrow flat tool lists and adopt nested-discovery patterns. Coupling the two means re-registering abilities every time the projection question is reopened.
+
+Decoupling them means registrations stay where they are while projections evolve underneath.
+
+## The three layers
+
+| Layer | Purpose | Examples |
+|---|---|---|
+| **Domain capability** | Stable, transport-neutral, permission-checked. The plugin's contract for "what can this do." | `myplugin/list-things`, `myplugin/update-thing`, `myplugin/get-overview` |
+| **Workflow** *(optional)* | Compositions of one or more abilities into a human- or agent-friendly task. | "Onboard a new merchant" = `verify-account` → `enable-feature` → `send-welcome` |
+| **Projection** | Consumer-specific view of the domain layer. Token-efficient compression for MCP, curated workflows for Command Palette, discoverable execution surface for REST, forms for admin UI, chainable commands for CLI. | An MCP projection might expose three meta-tools (`tools-list` / `tools-describe` / `tools-invoke`) over the same set of domain abilities that Command Palette shows flat. |
+
+The domain layer is what the registry holds. The projection layer is what each consumer sees. The workflow layer is optional and exists when a user-or-agent-meaningful task needs more than one ability.
+
+## The use-case-contract test
+
+A registered ability is a use-case contract — a natural-language shortcut to an action a human can already perform in the UI. REST endpoints and CLI commands are transport contracts; they expose plumbing. Abilities expose actions.
+
+Two consequences fall out of that framing:
+
+### 1. Inclusion test — "would a human ever do this in admin?"
+
+If yes, the operation is a candidate for an ability. If no, it stays an endpoint or a CLI command. Internal-only plumbing (cache invalidation, scheduler ticks, debug hooks) does not belong as an ability — even if it has a clean schema.
+
+The test is asymmetric: not everything an admin can do should be an ability either (rule 1 in `grouping-heuristic.md` covers grouping). But anything that isn't admin-meaningful almost certainly isn't ability-meaningful.
+
+### 2. Same code path as the UI
+
+If the ability mirrors a UI action, it must share the UI's permissions, validation, and business rules. The mechanism for keeping that promise is in `shared-core-service.md`. The framing is here: the ability is the use-case; UI / REST / CLI / MCP are thin adapters over it.
+
+## Implications
+
+### You may register abilities you do NOT expose via MCP
+
+The Abilities API is a WordPress core primitive. It establishes a formal contract for "run this code with these inputs under this permission check." MCP, REST, and Command Palette are exposure layers on top. You opt in.
+
+This means a plugin can register abilities that the Command Palette uses but MCP does not see, or vice versa. The registration is transport-neutral; exposure is per-consumer.
+
+### The same domain ability can appear in multiple projections with different shapes
+
+A small plugin might expose its three abilities flat in MCP and flat in Command Palette. A larger plugin might keep the same three domain abilities but project them through a nested-discovery facade for MCP (three meta-tools that traverse to the underlying registrations) while showing them flat in the Command Palette where token budget is not a constraint.
+
+The domain layer does not change. The projection layer does the consumer-specific work.
+
+### Token-efficiency tradeoffs live at the projection layer
+
+The choice between flat-with-full-schemas, semantic-grouping, single-tool-facade, and nested-discovery is a *projection* decision. Different shapes serve different consumer constraints. None of those shapes require changing what abilities are registered at the domain layer; they change how a consumer sees the registered set.
+
+This is what "pattern-agnostic" means in practice: the registration is one decision, the projection is another, and a redesign of the second does not invalidate the first.
+
+## Decision order
+
+When designing a plugin's ability surface, work in this order:
+
+1. **Domain first.** What can this plugin do, transport-neutrally? Use the inclusion test. Apply `grouping-heuristic.md` to pick the right granularity (semantic-intent over REST-atomization).
+2. **Projection second.** Which surfaces does each capability appear on, and in what shape? For most plugins under ~10 abilities, flat-in-every-projection works. For larger plugins, consider nested-discovery for MCP while keeping flat for Command Palette and REST.
+3. **Workflow third (only if needed).** Are there multi-ability tasks worth composing into a single user-or-agent-facing entry point? If yes, the workflow lives above the domain layer and references multiple abilities.
+
+Most plugins never need step 3.
+
+## Worked example — generic Notifications plugin
+
+A "Notifications" plugin manages a per-user notification feed. It supports listing, marking read, and dismissing all.
+
+### Domain layer
+
+Three abilities, registered once:
+
+- `notifications/list` — read; filter by `unread_only`, `since`, `category`.
+- `notifications/mark-read` — write; takes a `notification_id`.
+- `notifications/dismiss-all` — write; zero-arg.
+
+These are use-case contracts. They are transport-neutral. They do not assume MCP or Command Palette or REST.
+
+### Projection — MCP
+
+Small surface, flat works. The MCP projection exposes the three abilities as-is. No nested-discovery needed.
+
+### Projection — Command Palette
+
+The Command Palette projection adds a small workflow that wraps `notifications/list` with default filters (`unread_only=true`) and opens the admin notifications page on selection. The other two abilities remain visible as flat commands. The domain registration did not change; the Command Palette layer added the workflow on top.
+
+### Projection — REST
+
+The REST projection exposes only the read ability. Writes go through admin UI for security-review reasons. The domain registration is unchanged; the REST layer just doesn't expose two of the three.
+
+Same three registrations, three different projections, no re-registration as projection decisions evolve.
+
+## Escape hatch — tiny plugins
+
+For plugins with one to three admin-only abilities and no other surfaces, the projection layer is trivially the identity function. You don't need to *implement* layers; you need to *think* in layers so you don't paint yourself into a corner when the plugin grows or when MCP token budgets become a constraint.
+
+Concretely: name the abilities at the domain level (`myplugin/get-thing`, not `myplugin-mcp/get-thing`), keep the registration permission-checked and transport-neutral, and skip the projection layer until something needs it.
+
+## Related references
+
+- `grouping-heuristic.md` — within-domain decisions: how many abilities, where to put filters.
+- `shared-core-service.md` — the implementation mechanism for "same code path as the UI."
--- a/skills/wp-abilities-api/references/domain-vs-projection.md
+++ b/skills/wp-abilities-api/references/domain-vs-projection.md
@@ -31,11 +31,13 @@ A registered ability is a use-case contract — a natural-language shortcut to a
 
 Two consequences fall out of that framing:
 
-### 1. Inclusion test — "would a human ever do this in admin?"
+### 1. Inclusion test — "would a human intentionally do this through a supported UI or workflow?"
 
-If yes, the operation is a candidate for an ability. If no, it stays an endpoint or a CLI command. Internal-only plumbing (cache invalidation, scheduler ticks, debug hooks) does not belong as an ability — even if it has a clean schema.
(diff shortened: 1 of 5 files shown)


skills/wp-abilities-api/references/domain-vs-projection.md
```diff
@@ -0,0 +1,111 @@
...
+
+The domain layer is what the registry holds. The projection layer is what each consumer sees. The workflow layer is optional and exists when a user-or-agent-meaningful task needs more than one ability.
+
+## The use-case-contract test
+
+A registered ability is a use-case contract — a natural-language shortcut to an action a human can already perform in the UI. REST endpoints and CLI commands are transport contracts; they expose plumbing. Abilities expose actions.
+
+Two consequences fall out of that framing:
+
+### 1. Inclusion test — "would a human ever do this in admin?"
```
nerrad: Why is this framed as admin only? I think it's fair that abilities express functionality that other plugins might call as a part of a chain of actions (or the user/agent via various interfaces including CLI).

skills/wp-abilities-api/references/domain-vs-projection.md
```diff
@@ -0,0 +1,111 @@
...
+
+If yes, the operation is a candidate for an ability. If no, it stays an endpoint or a CLI command. Internal-only plumbing (cache invalidation, scheduler ticks, debug hooks) does not belong as an ability — even if it has a clean schema.
+
+The test is asymmetric: not everything an admin can do should be an ability either (rule 1 in `grouping-heuristic.md` covers grouping). But anything that isn't admin-meaningful almost certainly isn't ability-meaningful.
```
nerrad: I think this is prematurely specific (see previous comment). What about abilities scoped to a public authed/unauthed user. For example:

- `store/get-my-orders`, `store/get-order-status`
- `courses/list-my-courses`, `courses/get-progress`
- `appointments/list-available-slots`
- `events/get-my-tickets`, `events/get-available-tickets`
- `profile/update-public-profile`

...
Maybe good criteria would be "Would a human intentionally do this through a supported UI or workflow?"

skills/wp-abilities-api/references/grouping-heuristic.md
```diff
@@ -0,0 +1,91 @@
...
+
+REST-atomization would ship ~6 abilities (list, get, delete, update, bulk-update, count). Jetpack Forms instead registers **three**:
+
+- `jetpack-forms/get-responses` — list/search, with `status`, `is_unread`, `search`, `before`, `after`, `parent` in `input_schema`.
+- `jetpack-forms/update-response` — one write that covers status changes AND read-state toggles on a single response (semantically "modify a response").
+- `jetpack-forms/get-status-counts` — the dashboard summary ability. Zero-arg-friendly (only optional filters).
+
+Why three works: a user asking "show me spam responses from last week" uses one ability. An agent updating one response to `spam` uses one ability. The dashboard-style "how many unread?" uses one ability. The entire product surface fits in three tool-list entries.
+
+Canonical file: [`Automattic/jetpack/projects/packages/forms/src/abilities/class-forms-abilities.php`](https://github.com/Automattic/jetpack/blob/trunk/projects/packages/forms/src/abilities/class-forms-abilities.php).
```
nerrad: I don't think we should reference Jetpack forms explicitly here because it could get out of date. Would be better to use more generic examples.

skills/wp-abilities-api/references/shared-core-service.md
```diff
@@ -0,0 +1,151 @@
...
+
+The middle row is fine when the backing endpoint is light, read-only, and metric-neutral. The third row is what removes drift entirely; it is also more invasive because it usually means refactoring the existing REST controller to call the service rather than embed the logic.
+
+## The metric trap
+
+Most production REST endpoints in real plugins emit usage telemetry — analytics events, structured log emitters, custom event hooks. These metrics drive product decisions (which endpoints are hot, which are abandoned, which are slow).
+
+Routing an ability through `delegate_to_rest_controller` re-uses the REST controller's full code path, including its telemetry emission. Every agent invocation now counts as a UI-driven REST call. The metric is no longer a measure of human-driven usage; it is contaminated by agent traffic, and the contamination has no provenance label distinguishing the two.
```
nerrad: `delegate_to_rest_controller` assumes a real thing which may or not be the case outside of the context of this reference doc. So the guidance is mostly non-actionable. Either include the minimal delegation example to illustrate the pattern, or make sure this is phrased as something more conceptual (rather than concrete).

skills/wp-abilities-api/references/shared-core-service.md
```diff
@@ -0,0 +1,151 @@
...
+        return array_map( array( Things_Formatter::class, 'format' ), $rows );
+    }
+}
+
+// includes/admin/class-rest-things-controller.php
+class My_Plugin_REST_Things_Controller {
+    public function get_things( WP_REST_Request $request ) {
+        $rows = $this->service->list_things( $request->get_params() );
+
+        // Telemetry stays on the REST adapter — clean.
```
nerrad: Maybe?  I don't know that we need to explicitly include telemetry implementation in the examples? I could see scenarios where plugin authors might want to gather telemetry at different layers. Regardless that doesn't seem to be the important point of this guidance?

skills/wp-abilities-api/references/shared-core-service.md
```diff
@@ -0,0 +1,151 @@
...
+- **Read with telemetry on the REST handler** → extract a service. Don't contaminate metrics.
+- **Write of any kind** → extract a service. Drift is most damaging on writes (lost validation, missing audit hooks).
+- **No existing REST endpoint** → start at the service. The first ability you ship is also the right time to add the structure that a future REST endpoint will consume.
+
+The telemetry and write rules **override** the lighter "if the backing
+takes a `WP_REST_Request`, just delegate" heuristic. The signature test
+is sufficient only when telemetry is absent and the operation is a
+read. If the REST handler emits telemetry or the operation writes,
+extract a service even when delegating would otherwise be the easy
+path.
```
nerrad: As noted in earlier comment, I'm not sure why telemetry usage has to be emphasized here.

nerrad (changes requested):
One other thing to note generally. It looks like the registration guide for this skill is stale against current WordPress docs around the `permission_callback` and `meta.annotations` properties. Good opportunity to clean that up here too.

skills/wp-abilities-api/references/domain-vs-projection.md
rtio: Confirmed and broadened in `a18014f`. The inclusion-test subhead is now *"would a human intentionally do this through a supported UI or workflow?"* — that lens covers admin abilities (the original case), public-facing surfaces (storefront, account dashboard, course viewer, appointment booker), end-user self-service flows, and your chain-of-actions case where another plugin or an agent calls the operation as part of a workflow. Internal-only plumbing exclusion preserved (cache invalidation, scheduler ticks, debug snapshots, lifecycle bookkeeping).

skills/wp-abilities-api/references/domain-vs-projection.md
rtio: Same commit `a18014f`. Used the criterion you suggested almost verbatim and added explicit examples spanning both surfaces — `store/get-my-orders`, `events/list-available-tickets`, `profile/update-public-profile` from your list, alongside admin-side equivalents like `myplugin/list-pending-orders` and `myplugin/approve-submission`. The "anything that isn't admin-meaningful almost certainly isn't ability-meaningful" sentence that encoded the bias is gone.

skills/wp-abilities-api/references/grouping-heuristic.md
rtio: Replaced in `aea7a6a`. Worked example A is now a generic feedback / form-response plugin (`feedback/get-responses`, `feedback/update-response`, `feedback/get-status-counts`); the structural lessons are unchanged. The hardcoded canonical-file link to the Jetpack source is gone — that was the rot risk you flagged. The semantic-intent example in the table at line 13 also moved off `jetpack-forms/get-responses` to a generic equivalent.

skills/wp-abilities-api/references/shared-core-service.md
rtio: Confirmed — this would have looked like a fake function to anyone reading the doc in isolation, and the canonical helper of that name lives in a forthcoming PR that PR-44 can't link to without phantom-dependency. Two changes in `eaa75b6`:

- Every `delegate_to_rest_controller(...)` reference rephrased as prose: "the delegation pattern", "the delegate-to-REST-controller route".
- New subsection **"What the delegation pattern looks like"** with a self-contained 12-line PHP example showing the actual mechanism — construct `WP_REST_Request`, set query params, dispatch via `rest_do_request()`, handle the error case, return data. All five core symbols cited (`WP_REST_Request`, `rest_do_request`, `set_query_params`, `is_error`, `as_error`) verified against `wordpress-develop` trunk; `php -l` on the extracted block returns no syntax errors.

The reference is now actionable on its own without depending on any framework helper to be in place.

skills/wp-abilities-api/references/shared-core-service.md
rtio: This was over-indexed. Restructured in `1359274`:

- The H2 "The metric trap" is now "Side effects on the existing REST path" — telemetry is one of five bullets (alongside audit logs, custom-event hooks, email/notification dispatch, and cache / schedule / lock side effects).
- The `my_plugin_track_event(...)` calls in both worked-example PHP blocks are gone — replaced with `do_action( 'my_plugin/things_listed', $rows )`, which is a generic WP-idiomatic side effect that demonstrates the same point without baking a specific telemetry approach into the example. Plugin authors who track usage at a different layer can read past the bullet without feeling the example is presumptive.

Telemetry remains as the lead bullet (with the silent-failure-mode callout) because metric double-counting is the only side effect in the set where the system "works" while the dashboards drift — but it's no longer the centerpiece. Net: telemetry is now mentioned three times in the file, down from six.

skills/wp-abilities-api/references/shared-core-service.md
rtio: Same commit `1359274`. The "Read with no telemetry" / "Read with telemetry on the REST handler" rule-of-thumb bullets have been generalised to "Read with no side effects on the REST path" / "Read where the REST handler does more than data-fetch (audit, hooks, notifications, telemetry)". The closing override-paragraph below the bullets was rewritten in the same shape — "side-effect and write rules override the lighter delegate heuristic", with telemetry no longer at the centre of the framing.

rtio:
Hey Darren — round 3 just pushed. Three commits incorporating feedback that surfaced while I was reviewing the WooPayments pilot ([Automattic/woocommerce-payments#11679](https://github.com/Automattic/woocommerce-payments/pull/11679)) and your Woo Core MCP cleanup ([woocommerce/woocommerce#64425](https://github.com/woocommerce/woocommerce/pull/64425)):

- `e5a0cff` — teach `meta.mcp.public` and `meta.mcp.type` so abilities registered by following this skill are actually discoverable through the bundled MCP adapter.
- `86ed413` — escalate the three `meta.annotations` keys from "Optional" to "Strongly recommended" with reasoning, since the previous framing was licensing agents to skip them entirely.
- `19cbb6f` — add first-call REST bootstrap cost as a sixth side-effect category in `shared-core-service.md`, addressing the "what about calling the request class directly?" angle.

Total `+32 / -13` over the round-2 tip. The full round-3 disposition table, source-of-truth verifications, and what I deferred (a capability-scope nudge — your "permissions scope" hint admits more than one reading and I'd rather wait for specifics than guess) are in the updated PR body.

nerrad:
One generalized learning from using these skills to implement abilities across several WooCommerce extensions that seems worth folding into this strategy layer:

Abilities should model the domain/user intent first, not the transport shape. REST delegation is a useful implementation shortcut, but only when the REST path is a low-risk, read-only, side-effect-free data path. If the REST handler emits telemetry, hooks, notifications, audit logs, cache writes, or expensive bootstrap behavior, the ability should call shared domain/service logic instead.

This PR already points in that direction. I’d suggest making the “delegate to REST” path read as conditional rather than broadly default, and explicitly noting that plugins may choose stricter implementation rules.

skills/wp-abilities-api/references/domain-vs-projection.md
```diff
@@ -0,0 +1,113 @@
...
+
+Decoupling them means registrations stay where they are while projections evolve underneath.
+
+## The three layers
+
+| Layer | Purpose | Examples |
+|---|---|---|
+| **Domain capability** | Stable, transport-neutral, permission-checked. The plugin's contract for "what can this do." | `myplugin/list-things`, `myplugin/update-thing`, `myplugin/get-overview` |
+| **Workflow** *(optional)* | Compositions of one or more abilities into a human- or agent-friendly task. | "Onboard a new merchant" = `verify-account` → `enable-feature` → `send-welcome` |
+| **Projection** | Consumer-specific view of the domain layer. Token-efficient compression for MCP, curated workflows for Command Palette, discoverable execution surface for REST, forms for admin UI, chainable commands for CLI. | An MCP projection might expose three meta-tools (`tools-list` / `tools-describe` / `tools-invoke`) over the same set of domain abilities that Command Palette shows flat. |
```
nerrad: > An MCP projection might expose three meta-tools (`tools-list` / `tools-describe` / `tools-invoke`) over the same set of domain abilities that Command Palette shows flat.

There is the existing example of the WordPress/mcp-adapter that might be better to lean into here?
 
- `mcp-adapter-discover-abilities`
- `mcp-adapter-execute-ability`
- `mcp-adapter-get-ability-info`

rtio: @nerrad — done in `25530c6`. The cell now names the real adapter abilities: `mcp-adapter/discover-abilities`, `mcp-adapter/execute-ability`, `mcp-adapter/get-ability-info`. Verified against `WordPress/mcp-adapter/includes/Abilities/*.php` trunk. Drops the "meta-tools" misnomer in passing — `tools/list` / `tools/describe` / `tools/invoke` are MCP protocol primitives, not meta-tools. Command Palette contrast preserved.

rtio:
@nerrad — done in `479c48c`. Reframed `shared-core-service.md` in four places: intro names the shared service as the default; three-shapes table verdicts flip ("Delegate" → conditional shortcut with the three conditions enumerated; "Extract a service" → default); rule-of-thumb reordered so service-mandatory cases come first; new `Plugin-family override` section noting payments / subscriptions / own-MCP-transport plugins routinely tighten further. Nothing from rounds 2–3 undone — framing weight changed, not surface area. Round-4 disposition table is in the PR body.

rtio:
@nerrad — round 4 is up (mcp-adapter projection example anchored; REST delegation reframed as a conditional shortcut). Ready for formal approve when convenient.

enejb (approved):
These rules make sense to me and a great improvement.

nerrad (approved):

JasonTheAdams (approved):

Issue #21, opened 2026-02-14
New Skill Proposal: wp-plugins-directory-guidelines


davidperezgar:
Add a new skill called **`wp-plugins-guidelines`** that provides AI coding assistants with the complete WordPress.org Plugin Directory guidelines and GPL licensing requirements. This ensures that AI-generated plugins comply with directory submission rules from the start, avoiding common rejection reasons.

## Problem

AI coding assistants frequently generate WordPress plugins that violate Plugin Directory guidelines — for example:

- Including trialware or time-locked features (Guideline 5)
- Tracking users without consent (Guideline 7)
- Sending executable code via third-party systems (Guideline 8)
- Bundling libraries already included in WordPress core (Guideline 13)
- Embedding external links/credits without user permission (Guideline 10)
- Using incorrect or incompatible licenses (Guideline 1 — GPL compatibility)
- Hijacking the admin dashboard with intrusive notices (Guideline 11)

These issues lead to plugin rejections during review, wasted developer time, and potential security/privacy concerns in production sites.

### SKILL.md Scope

The `SKILL.md` would include:

- **When to use**: Triggered when creating a new plugin for WordPress.org directory submission, preparing a plugin for review, or when the user mentions "plugin guidelines", "directory submission", "GPL compliance", etc.
- **Procedure**: A checklist the agent follows to validate the plugin against all 18 guidelines before considering the plugin ready for submission.
- **Key checks**:
  - GPL-compatible license declared in plugin header and `readme.txt`
  - No tracking without explicit user consent
  - No remote executable code loading
  - WordPress default libraries used (jQuery, React, etc. from core, not bundled)
  - No trialware patterns
  - No admin dashboard hijacking (aggressive notices, banners, redirects)
  - Proper readme format (no spam, correct stable tag, etc.)
  - Version numbering follows WordPress conventions
  - Complete, functional plugin (no stubs/placeholders for submission)
  - Trademarks and copyrights respected in naming
- **Verification**: Steps to confirm compliance before submission.
- **Failure modes**: Common rejection patterns and how to fix them.

### References

- The full text of the [[Detailed Plugin Guidelines](https://developer.wordpress.org/plugins/wordpress-org/detailed-plugin-guidelines/)](https://developer.wordpress.org/plugins/wordpress-org/detailed-plugin-guidelines/) (all 18 guidelines with their explanations), so the agent has complete context without needing to fetch external URLs.
[...]

PR #45, opened 2026-04-30
wp-abilities-api: add four mechanics-layer references (plugin-family-patterns, delegate-helper-pattern, error-code-vocabulary, input-schema-gotchas)

rtio:
Adds four mechanics-layer reference docs to `wp-abilities-api/` plus a six-line cross-link block in `SKILL.md`. Stacks on merged #44 (now in `trunk` as `26aa4e8`).

## What lands

| File | Purpose |
|---|---|
| `references/plugin-family-patterns.md` | Two call-shape patterns (shared API client vs zero-arg controllers) — implementation *mechanics* only. Family-specific *registration* conventions (loader, category, MCP defaults, error-code prefix) explicitly belong in plugin-family overlays. Service-extraction is named as the third option outside the A/B dichotomy, coherent with #44 round 4. |
| `references/delegate-helper-pattern.md` | Canonical `delegate_to_rest_controller` helper. Opens by gating entry on the three delegation preconditions (read-only handler, read op, predominantly inside REST contexts); for everything else, points back at `shared-core-service.md`. Three guards, dual response-shape unwrap, when NOT to use. |
| `references/error-code-vocabulary.md` | Standardized `WP_Error` codes (`<plugin>_not_initialized`, `<plugin>_missing_<field>`, `<plugin>_invalid_<field>`, `<plugin>_<resource>_data_unavailable`, `ability_invalid_input`) so agents can build retry logic against codes, not message strings. |
| `references/input-schema-gotchas.md` | Four runtime surprises: schema-default non-injection, pagination key drift between abilities exposing `per_page` and backings reading `pagesize`, PHP `empty()` false-rejecting `"0"` and `0` as legitimate IDs, and direct-vs-indirect invocation strictness. |
| `wp-abilities-api/SKILL.md` | Six lines: step-4 cross-links + a failure-mode pointer. |

## Verification at tip

- WP core symbols verified against `WordPress/wordpress-develop` trunk.
- No plugin-family-specific names in worked examples after round 2.

## Status

Round 2 (current tip). Round 2 addressed @nerrad's 2026-05-21 inline comment at `#discussion_r3277876461` asking for an overlay-vs-portable framing in `plugin-family-patterns.md:3`, plus the related portability cleanup across the other refs:

- `acfea3c` — rewrites `plugin-family-patterns.md`'s intro on the overlay axis; drops `("Woo-family")` / `("Jetpack-family")` parentheticals on Pattern A/B; genericizes both worked examples (WooPayments class chain → generic SaaS client; Jetpack Forms file link → generic CPT pattern) and the family-named "Picking quickly" row.
- `5eb7f72` — same posture applied to `error-code-vocabulary.md` (drops `woopayments` / `jetpack_forms` and the WooPayments house-style callout).
- `e16d580` — re-anchors `delegate-helper-pattern.md`'s opening to the conditional-shortcut framing from #44 round 4.

This reference is now portable mechanics only; family-specific registration conventions belong in the WooCommerce-extension overlay (`Automattic/payments-ai-tools` PR #21).

Rebased onto `trunk` after #44 merged on 2026-05-26 (squash commit `26aa4e8`); the diff now shows only this PR's 12 mechanics-layer commits.

### DIFF
--- a/skills/wp-abilities-api/references/delegate-helper-pattern.md
+++ b/skills/wp-abilities-api/references/delegate-helper-pattern.md
@@ -1,0 +1,215 @@
+# Delegate helper pattern
+
+When an ability's execute callback needs to hand work to an existing REST controller instead of duplicating business logic, extract a `delegate_to_rest_controller` helper. This reference documents the canonical signature, the three guards every implementation needs, the dual response-shape unwrap, and when NOT to use the helper at all.
+
+Read `plugin-family-patterns.md` first — the helper's exact shape depends on whether your plugin follows the shared-API-client pattern or the zero-arg-controllers pattern.
+
+## Why this helper exists
+
+List-style read abilities typically repeat the same four steps in every execute callback:
+
+1. Validate the plugin is initialized (class exists + dependencies resolved).
+2. Build a `WP_REST_Request` and copy the ability's `$input` onto it as params.
+3. Instantiate the backing controller and invoke the named method.
+4. Unwrap the response (controllers return either a raw array or a `WP_REST_Response`).
+
+Four abilities × this repetition = the boilerplate dominates the callback body. Extracting a helper keeps each execute callback a 3–4 line function that reads like pseudocode.
+
+## When to extract
+
+After the **second** list-style ability lands. Before the third ability gets added, refactor the duplicated code out of the first two execute callbacks into a private static helper. Convert subsequent abilities to call the helper as you add them.
+
+If the plugin only ever ships one or two abilities, inline the code. The helper's payoff is at 3+ callers.
+
+## Canonical signature (Pattern A — shared-API-client)
+
+```php
+/**
+ * Delegate an ability's execute callback to a REST controller.
+ *
+ * Builds a WP_REST_Request from the ability's input, instantiates the
+ * backing controller with the plugin's shared API client, invokes the
+ * named method, and normalizes the return so callers always see
+ * array|\WP_Error. Controllers in this plugin return either a
+ * WP_REST_Response or a raw array; the helper unwraps both shapes.
+ *
+ * Not used by zero-arg abilities — those call their controller directly
+ * so we don't synthesize a WP_REST_Request just to discard it.
+ *
+ * @param string     $controller_class Fully-qualified controller class (no leading backslash).
+ * @param string     $method           Controller method to invoke on the built request.
+ * @param string     $route            REST route string used when constructing WP_REST_Request.
+ * @param array|null $input            Ability input; each key/value becomes a request param.
+ * @param string     $http_method      HTTP method for WP_REST_Request. Defaults to 'GET'; writes pass 'POST'.
+ * @return array|\WP_Error             Response payload as an array, or WP_Error on failure.
+ */
+private static function delegate_to_rest_controller(
+    string $controller_class,
+    string $method,
+    string $route,
+    ?array $input,
+    string $http_method = 'GET'
+) {
+    $fqcn = '\\' . $controller_class;
+
+    // Guard 1: controller class is loaded.
+    if ( ! class_exists( $fqcn ) ) {
+        return new \WP_Error(
+            '<plugin>_not_initialized',
+            __( '<Plugin> is not initialized.', '<text-domain>' )
+        );
+    }
+
+    // Guard 2: shared API client is available.
+    $api_client = null;
+    if ( class_exists( '\<Plugin_Main_Class>' ) && method_exists( '\<Plugin_Main_Class>', 'get_<client_accessor>' ) ) {
+        $api_client = \<Plugin_Main_Class>::get_<client_accessor>();
+    }
+    if ( null === $api_client ) {
+        return new \WP_Error(
+            '<plugin>_not_initialized',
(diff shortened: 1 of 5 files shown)


skills/wp-abilities-api/references/plugin-family-patterns.md
```diff
@@ -0,0 +1,229 @@
+# Plugin-family patterns
+
+When you adopt the Abilities API inside an existing plugin, the plugin's architecture dictates HOW your execute callbacks call the existing business logic. Two patterns cover most real-world WordPress plugins. Pick one up front — the choice ripples through your delegate helper, your tests, and your error codes.
```
nerrad: Could this intro clarify that "plugin family" guidance covers shared implementation mechanics, but family-specific registration conventions may still live in separate overlays?

For example, WooCommerce extensions may share loader/category/MCP conventions that should not become defaults for every WordPress plugin. This reference can stay portable while telling implementers to apply any plugin-family overlay before scaffolding registration.

rtio:
@nerrad — ready for first review when you have a window. Round 2 just landed addressing your inline at #discussion_r3277876461 (overlay-vs-portable framing in `plugin-family-patterns.md`) plus the related portability cleanup across the other refs. Stacks on #44. Slim pass deferred to post-merge per the project's substantive-pre-merge posture.

skills/wp-abilities-api/references/plugin-family-patterns.md
rtio: @nerrad — done in `acfea3c` + `5eb7f72` + `e16d580`. The first rewrites `plugin-family-patterns.md`'s intro on the overlay-vs-portable axis, drops the `("Woo-family")` / `("Jetpack-family")` parentheticals on Pattern A/B, genericizes both worked examples (WooPayments class chain → generic SaaS client; Jetpack Forms file link → generic CPT pattern), and replaces the family-named "Picking quickly" row with a shape-based signal. The second does the same to `error-code-vocabulary.md`. The third re-anchors `delegate-helper-pattern.md` to the conditional-shortcut framing from #44 round 4. Reference is now portable mechanics only; family-specific registration conventions belong in the WooCommerce overlay.

skills/wp-abilities-api/references/delegate-helper-pattern.md
```diff
@@ -0,0 +1,241 @@
...
+
+Four abilities × this repetition = the boilerplate dominates the callback body. Extracting a helper keeps each execute callback a 3–4 line function that reads like pseudocode.
+
+## When to extract
+
+After the **second** list-style ability lands. Before the third ability gets added, refactor the duplicated code out of the first two execute callbacks into a private static helper. Convert subsequent abilities to call the helper as you add them.
+
+If the plugin only ever ships one or two abilities, inline the code. The helper's payoff is at 3+ callers.
+
+## Canonical signature (Pattern A — shared-API-client)
```
nerrad: Why does this need "Pattern A - shared-API-client" - it seems like an artifact from when different patterns were being evaluated. I'm also wary that using the phrase "Canonical signature" here will too strongly weight this as the preferred pattern.

nerrad (commented):
As I've mentioned elsewhere, I'm wary that getting too specific in these skill references skews models too much toward the specific pattern rather than solving the intent via principles. There's risk that the code examples in the skill get out of date and actually end up making things worse. Especially if this was generated _by a current generation model_. 

I'm not going to block the PR on this point, but I think it's an important consideration you should have in mind.

skills/wp-abilities-api/references/delegate-helper-pattern.md
rtio: Both qualifiers dropped. The H2 is now `## Helper shape`, the intro reframes the example as "one helper shape that works — adapt the exact signature to the plugin you're working in," and the body references that named Pattern A/B plugins now describe the situation ("when the controller takes a shared API client" / "when the controller takes no constructor args"). Pushed in `95b9d41`. Same sweep applied to `plugin-family-patterns.md` H2s in `d0bf7c7` for coherence.

rtio:
On the cross-cutting "too specific / risks capping models / risks going stale" — fair, and applied where the labels were doing more weighting than work:

- `#45`: dropped "Canonical signature (Pattern A — shared-API-client)" → "Helper shape"; renamed "Pattern A/B" H2s in `plugin-family-patterns.md` to "Shape: shared API client" / "Shape: zero-arg controllers"; reframed framing prose. `95b9d41`, `d0bf7c7`.
- `#47`: same one-word swap on `static-enumeration.md`'s "Canonical registration shape" → "Typical". `0e0e9e9`.

Kept the substance: the dichotomy in `plugin-family-patterns.md`, the three guards in `delegate-helper-pattern.md`, the dual response-shape unwrap. Those describe real WP idiosyncrasies a stronger model still can't infer from API docs, and the skill demonstrably works in practice — you used it to ship the WC overlay. Reshaping further would risk breaking what's verified working.

`#46` left as-is — its content is schema fields rather than prescriptive code, no analogous labels to soften.

nerrad (approved):

rtio:
Closing in favor of https://github.com/WordPress/agent-skills/pull/47

PR #46, opened 2026-04-30
wp-abilities-audit: add new skill for surveying a plugin's REST surface and producing a structured audit doc

rtio:
Adds the `wp-abilities-audit` skill — surveys a plugin's REST surface and produces a structured audit document proposing Abilities API registrations. The audit doc is a planning artifact: readable by humans, parseable by tooling. Stacks on #45 (merged #44 is in `trunk` as `26aa4e8`).

## What lands

| File | Purpose |
|---|---|
| `skills/wp-abilities-audit/SKILL.md` | 7-step procedure: enumerate controllers, extract backing fields, confirm capability gates, propose abilities via semantic-intent grouping, surface gaps, write the doc, optionally designate a reference ability. |
| `references/audit-schema.md` | Canonical schema. Top-level fields (plugin, repo, branch, dates, capability_gate, plus optional plugin_family). Per-ability fields include the three required implementation-readiness fields (`use_case_fit`, `side_effects`, `seed_data_needs`) plus `backing.kind` (`rest_controller` / `service` / `helper` / `data_store`) and `permission.source` (`rest_controller` / `admin_action` / `service` / `domain_policy` / `post_type_map` / `none`). `side_effects: []` is the load-bearing fact that unlocks the conditional delegation shortcut in `shared-core-service.md`; a non-empty array selects the shared-service shape. Two legal `capability_gate` forms (single string vs structured `{read, write}`). `backing: null` semantics. WARN-not-FAIL posture for audits authored before the schema bumps. Copy-pasteable minimal example. |
| `references/capability-gate-tracing.md` | Procedure for tracing the *permission source* (REST callback / admin action / service method / domain policy / post-type cap map) for each proposed behavior, not just the REST route. Mechanism A (direct: base controller calls `current_user_can()` once) and Mechanism B (post-type-backed: controllers extend `WP_REST_Posts_Controller`-family, permission resolves dynamically via core's `map_meta_cap()`). Tracing recipes and YAML representation for each. |
| `references/controller-enumeration.md` | Two enumeration paths — glob for standard layouts, grep as universal fallback — with guidance on monorepos, inherited routes, and the exhaustiveness rule. |
| `eval/scenarios/wp-abilities-audit.json` | Eval scenario covering a non-standard REST layout with post-type-backed capabilities and a missing-backing gap; success criteria assert the three implementation-readiness fields get populated. |

## Verification at tip

- All cross-references resolve within the #45/#46 stack on top of merged #44.

## Status

Round 4 (current tip). Iteration:

- **Round 2** (`34cfb58`) — added three implementation-readiness fields (`use_case_fit`, `side_effects`, `seed_data_needs`) per @nerrad's 2026-05-21 generalized learning.
- **Round 3** (`ac09d3b`) — tied `side_effects` to the Shape 2 vs Shape 3 decision in `shared-core-service.md`; added the backwards-compat WARN bullet; genericized the `seed_data_needs` example; updated the eval scenario.
- **Round 4** (`238d772` + `1612600` + `e597aaf`) — addressed @nerrad's three inline comments from 2026-05-21 00:18 UTC. `238d772` adds `backing.kind` and `permission.source` enum fields (closes the bullet-1 ask in `#discussion_r3277895737` that bullets 2 and 3 closed in round 2). `1612600` adds the use-case sanity check to the SKILL.md procedure before populating ability fields (closes `#discussion_r3277895795`). `e597aaf` rewrites the permission-tracing procedure to frame drift against plugin behavior rather than REST route, with the new `permission.source` enum as the recording field (closes `#discussion_r3277895951`).

Rebased after #44 merged on 2026-05-26; once #45 merges, this PR's diff narrows to just the 11 audit-skill commits.

### DIFF
--- a/skills/wp-abilities-audit/SKILL.md
+++ b/skills/wp-abilities-audit/SKILL.md
@@ -91,7 +91,15 @@ variants into 1.
 
 For each proposed ability, fill in every field in the `proposed_abilities`
 schema: `name`, `intent`, `backing`, `permission`, `return_type`, `effort`
-(S/M/L), `annotations` (readonly/destructive/idempotent), `notes`, `risks`.
+(S/M/L), `annotations` (readonly/destructive/idempotent), `notes`, `risks`,
+`use_case_fit`, `side_effects`, `seed_data_needs`.
+
+The last three are the implementation-readiness facts the implementer
+and the verify-mode tooling both need: which human/agent workflow this
+ability serves (`use_case_fit`), what the backing path emits on every
+call (`side_effects` — empty array is a fact, not a missing value), and
+what representative data must exist in the test environment for the
+ability to execute through the public boundary (`seed_data_needs`).
 
 ### 5. Surface gaps and deferred items
 
--- a/skills/wp-abilities-audit/SKILL.md
+++ b/skills/wp-abilities-audit/SKILL.md
@@ -89,10 +89,22 @@ re-derive the rules here. Short version: one ability per real-world question
 or state transition, with filter parameters in `input_schema` collapsing N
 variants into 1.
 
-For each proposed ability, fill in every field in the `proposed_abilities`
-schema: `name`, `intent`, `backing`, `permission`, `return_type`, `effort`
-(S/M/L), `annotations` (readonly/destructive/idempotent), `notes`, `risks`,
-`use_case_fit`, `side_effects`, `seed_data_needs`.
+**Apply the use-case sanity check before populating any candidate.** Per
+`../wp-abilities-api/references/domain-vs-projection.md`'s use-case-contract
+test: would a human or agent intentionally perform this behavior through a
+supported plugin workflow? If yes, the candidate is a real ability —
+proceed to fill in fields. If no, the route is internal transport plumbing
+(cache invalidation, scheduler ticks, bookkeeping endpoints, debug
+introspection) — keep it in the Controller Inventory section for
+completeness, but do NOT promote it to `proposed_abilities`. The route may
+be useful to inventory; the proposed ability must represent a real
+user/operator question or action.
+
+For each proposed ability that passes the sanity check, fill in every
+field in the `proposed_abilities` schema: `name`, `intent`, `backing`,
+`permission`, `return_type`, `effort` (S/M/L), `annotations`
+(readonly/destructive/idempotent), `notes`, `risks`, `use_case_fit`,
+`side_effects`, `seed_data_needs`.
 
 The last three are the implementation-readiness facts the implementer
 and the verify-mode tooling both need: which human/agent workflow this
--- a/skills/wp-abilities-audit/SKILL.md
+++ b/skills/wp-abilities-audit/SKILL.md
@@ -1,0 +1,178 @@
+---
+name: wp-abilities-audit
+description: "Audit a WordPress plugin's REST surface and produce a standardized audit document proposing Abilities API registrations. Produces a markdown doc with a YAML schema and prose sections that humans and agents can both consume when planning a registration rollout. Works on any WP plugin."
+compatibility: "Targets WordPress 6.9+ (PHP 7.2.24+). Filesystem-based agent with bash + node. Requires access to the plugin checkout; some workflows benefit from WP-CLI but don't require it."
+---
+
+# WP Abilities Audit
+
+Produce a standardized audit document for a WordPress plugin's REST surface,
+proposing a set of Abilities API registrations grouped by semantic intent. The
+audit doc is a planning artifact for implementers — humans, agents, or both —
+that captures the controller inventory, capability gates, and proposed
+ability shapes in a structured form. A reviewer reading the doc can scope the
+work without re-deriving the survey.
+
+This skill works on any plugin that exposes a REST surface. Plugin
+classification (for purposes of the optional `plugin_family` annotation) is
+the user's call; the workflow itself is plugin-agnostic.
+
+## When to use
+
+- The task is "register Abilities API abilities for a WP plugin" and no audit
+  doc exists yet.
+- Planning participation in a multi-plugin abilities rollout and need a
+  shareable, standardized audit artifact.
+- Pre-flight checking a plugin's agent-readiness before implementing abilities.
+- A PM or non-implementer wants to scope the work before engineering picks it up.
+
+## Inputs required
+
+1. **Plugin checkout path** — working tree of the plugin to audit.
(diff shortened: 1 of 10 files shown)


skills/wp-abilities-audit/references/audit-schema.md
```diff
@@ -0,0 +1,267 @@
...
+|---|---|---|
+| `name` | string | Kebab-case `<plugin-slug>/<ability>`. |
+| `intent` | string | One sentence, user-question framed. |
+| `backing` | object or `null` | See below. `null` marks an ability with no backing endpoint (a known gap). |
+| `permission` | object or `null` | See below. `null` when `backing` is null. |
+| `return_type` | string | Short description (e.g. `WP_REST_Response (wrapping array)`). Hint-only; not machine-parsed. |
+| `effort` | enum | `S`, `M`, or `L`. |
+| `annotations` | object | `{ readonly: bool, destructive: bool, idempotent: bool }`. All three required. |
+| `notes` | array of strings | Implementer-facing detail (filter params, edge cases, alternative backings). |
+| `risks` | array of strings | Anything the implementer must handle (missing idempotency key, two-phase behavior, `permission_callback => '__return_true'` at the REST layer that must not copy into the ability, etc.). |
```
nerrad, line 83: Could the schema preserve a bit more implementation-readiness context here?

While using these skills to implement abilities across several WooCommerce extensions, the useful audit facts were not only “which REST route backs this?” but also:
- the backing implementation path the ability should use or inspect: REST controller, service, helper, data store, or no backing
- side effects on that path: telemetry, hooks, notifications, audit logs, cache writes
- representative seed data needed to verify execution later

Some of this could live in `notes` / `risks`, but it may be worth naming these explicitly so audits reliably preserve the facts that downstream implementation and verification need.

skills/wp-abilities-audit/SKILL.md
```diff
@@ -0,0 +1,178 @@
...
+Note explicitly whether read and write gates differ: compound gates are
+represented as a `{read, write}` object, not a single string.
+
+### 4. Propose abilities using semantic-intent grouping
+
+Do NOT atomize one ability per HTTP method. Apply the semantic-intent grouping
+heuristic — it's the only grouping rule this skill uses.
+
+Read `../wp-abilities-api/references/grouping-heuristic.md` now — do NOT
+re-derive the rules here. Short version: one ability per real-world question
```
nerrad, line 88: Could this section add a small use-case sanity check before proposing an ability?

A helpful rule from using these skills to implement abilities across several WooCommerce extensions: if no human would intentionally perform this behavior through a supported plugin workflow, it is probably internal transport plumbing rather than an ability. The route may still be useful to inventory, but the proposed ability should represent a real user/operator question or action.

skills/wp-abilities-audit/references/capability-gate-tracing.md
```diff
@@ -0,0 +1,171 @@
...
+- YAML comments after the string are silently dropped by strict parsers, so
+  provenance gets lost.
+- The `{read, write}` object form is machine-parseable and carries
+  `confirmed` and `verified_at` in-band.
+
+Prefer the structured form for any new audit.
+
+## Procedure — trace every route you'll back
+
+For each proposed ability, walk the chain once:
```
nerrad, line 157: Could this tracing step frame permission drift against the relevant plugin behavior, not only the REST route?

The ability permission should match the plugin’s intended gate for the proposed behavior. Often that is the REST controller permission callback, but in some plugins the relevant source of truth may be an admin UI action, service/helper method, domain policy, or post-type capability map. The audit should preserve where that permission source came from and whether the proposed ability could drift from it.

skills/wp-abilities-audit/references/audit-schema.md
rtio: @nerrad — done across two rounds. Bullets 2 (side effects) and 3 (seed data) landed earlier today as required schema fields (`side_effects`, `seed_data_needs`) in `34cfb58`. Bullet 1 (backing implementation path) landed just now in `238d772` as a new `backing.kind` enum: `rest_controller` / `service` / `helper` / `data_store`, optional with default `rest_controller` for backwards compat. REST-specific fields (`method`, `route`, `route_registration_line`) become not-applicable for non-REST kinds. The minimal-valid-example now demonstrates `kind: service` on the second proposed ability so the non-REST shape is concrete, not hypothetical. Why this matters in the schema rather than `notes` / `risks`: the kind drives the downstream Shape 2 vs Shape 3 decision per `shared-core-service.md` — only `rest_controller` is a candidate for the conditional delegation shortcut; the other kinds select the shared-service shape from the start.

skills/wp-abilities-audit/SKILL.md
rtio: @nerrad — done in `1612600`. Added the sanity-check paragraph in step 4 of `SKILL.md`, between the semantic-intent grouping reference and the "fill in every field" instruction. The check reads: would a human or agent intentionally perform this behavior through a supported plugin workflow? If yes → real ability, populate fields. If no → internal transport plumbing (cache invalidation, scheduler ticks, bookkeeping endpoints, debug introspection) — keep it in the Controller Inventory section for completeness, do NOT promote to `proposed_abilities`. The follow-on instruction now reads "For each proposed ability that passes the sanity check, fill in every field..." so the gate is explicit at the field-population step. Cross-linked to `domain-vs-projection.md`'s use-case-contract test.

skills/wp-abilities-audit/references/capability-gate-tracing.md
rtio: @nerrad — done across two commits. `238d772` adds a `permission.source` enum field to the schema (`rest_controller` / `admin_action` / `service` / `domain_policy` / `post_type_map` / `none`, optional with default `rest_controller`) so the audit preserves *where* the permission canonically lives, not just *what* it resolves to. `e597aaf` rewrites the tracing procedure: opening reframed from "trace every route you'll back" to "trace the permission source for each proposed behavior", with step 1 explicitly looking for canonical permission sources beyond the REST callback (admin handlers, service methods, domain policies, post-type cap maps) when the REST callback is `'__return_true'`, fully delegating, or otherwise not the source of truth. Step 6 (new) requires a risk note when the canonical source diverges from the REST callback. Step 7 (renumbered) hoists to `capability_gate` only when every behavior resolves to the same cap *at the same source* — divergence in either cap OR source belongs in Notes and Surprises.

nerrad (approved):

rtio:
Closing in favor of https://github.com/WordPress/agent-skills/pull/47

PR #47, opened 2026-04-30
wp-abilities-verify: add new skill for verifying Abilities API registrations including adversarial readonly-but-writes detection

rtio:
Adds the `wp-abilities-verify` skill — checks a plugin's Abilities API registrations against the structural and behavioral promises they declare. The centerpiece is **adversarial annotation correctness**: a `readonly: true` ability whose execute callback actually writes (via `$wpdb->update`, `update_option`, `wp_insert_*`, a non-GET delegate, etc.) is the bug class that unit tests don't catch — the mock looks just like the real writer. Stacks on #46 (merged #44 is in `trunk` as `26aa4e8`).

## What lands

| File | Purpose |
|---|---|
| `skills/wp-abilities-verify/SKILL.md` | Two modes (static, runtime). Seven-step procedure: validate audit doc → enumerate statically → enumerate at runtime (runtime mode) → annotation correctness → permission roundtrip → schema lints → error-code vocabulary. Static-mode blind spots called out so a static PASS isn't read as "verified write-free." |
| `references/static-enumeration.md` | Find `wp_register_ability(` calls (multi-line via `rg --multiline --pcre2`), extract names + annotation blocks, follow `execute_callback` to its body. |
| `references/annotation-correctness.md` | The adversarial core. Three-claims table anchored in the `idempotent` annotation's docblock in `class-wp-ability.php` and `WP_REST_Abilities_V1_Run_Controller::validate_request_method()`. Common write patterns as a starting set; blind spots (indirected service writes, hooks-as-writes, variable-built HTTP methods, tautological caps); the `// verify-ignore` suppression mechanism. |
| `references/schema-lints.md` | Six lint principles applied to each ability's `input_schema`. |
| `references/permission-roundtrip.md` | What `permission_callback` actually receives (validated input value, never a `WP_REST_Request` — the antipattern ships in plugins copied from REST controllers without translation). Six callback shapes (A-F) classified OK / WARN / FAIL. Runnable wp-cli block exercising the gate against anon / subscriber / admin contexts. |
| `references/runtime-harness.md` | Six wp-cli checks against a booted WordPress: name match, annotation read-back, read execute (split into Level 1 smoke / Level 2 high-confidence using `seed_data_needs` from the audit), write input validation, permission gate, idempotency twin-invocation heuristic. All execute-checks go through `wp_get_ability( $name )->execute( ... )` — verification through the public boundary by design. |
| `references/audit-schema-validation.md` | Delegates to the canonical schema in `wp-abilities-audit/references/audit-schema.md`; picks up #46's three new required fields automatically post-rebase. |
| `eval/scenarios/wp-abilities-verify.json` | Scenario where one read ability accidentally calls `$wpdb->update`; success criteria require FAIL with file + line evidence. |

## Verification at tip

- WP core symbols (`WP_Ability::*`, `WP_REST_Abilities_V1_Run_Controller::validate_request_method`, `wp_register_ability`, `wp_get_ability`, `wp_get_abilities`, etc.) verified against `WordPress/wordpress-develop` trunk.

## Status

Round 4 (current tip). Iteration:

- **Round 2** (`c5f962a..d8ea032`) — five focused fixes against @nerrad's review: `idempotent` wording aligned to core's docblock; `(object) array()` accepted in Lint 5; mutator-verb regex broadened; multi-line `wp_register_ability` handled; README updated.
- **Round 3** (`dad0196..f0070f6`) — ~650 lines net cut per @nerrad's "ship lighter; iterate from real usage" guidance. References now read as principles + a few examples + agent-derived patterns rather than exhaustive catalogues. Runnable harnesses preserved as concrete runbooks.
- **Round 4** (`b6e1bb4`) — addresses @nerrad's `#discussion_r3277906278` on `runtime-harness.md:127`. Splits Check 3 into Level 1 — smoke execution (synthetic inputs, prior content kept verbatim) and Level 2 — high-confidence verification (representative seeded data using `seed_data_needs` from the audit; asserts output shape AND privacy contract). When `seed_data_needs: null`, Level 2 reports `pending (ask the implementer)` — dovetails with #46's WARN-not-FAIL posture. Threaded reply posted.

Post-rebase follow-up (when #46 lands): one count update in `audit-schema-validation.md` ("All 7 required fields present" → generic, since the schema now has 10).

Rebased after #44 merged on 2026-05-26; once #45 and #46 merge, this PR's diff narrows to just the 25 verify-skill commits.

### DIFF
--- a/eval/scenarios/wp-abilities-verify.json
+++ b/eval/scenarios/wp-abilities-verify.json
@@ -1,0 +1,26 @@
+{
+  "name": "Verify abilities registrations and catch a readonly-but-writes violation",
+  "skills": ["wordpress-router", "wp-abilities-verify", "wp-abilities-audit", "wp-abilities-api", "wp-project-triage"],
+  "query": "I have a WordPress plugin checkout with 7 registered abilities — 6 read abilities annotated readonly: true and 1 write ability annotated readonly: false, destructive: true. One of the six read abilities accidentally calls $wpdb->update(...) inside its execute callback after a recent refactor. I also have an audit doc at ~/vault/plans/2026-04-20-abilities-audit-<plugin>.md that declared the expected annotations before the refactor. Run wp-abilities-verify in static mode against the plugin checkout and write the structured report to ~/vault/plans/2026-04-20-verify-<plugin>.md.",
+  "expected_behavior": [
+    "Step 1: Route to wp-abilities-verify because the task is 'verify abilities registrations'",
+    "Step 2: Validate the audit doc against audit-schema-validation.md (required fields, at-most-one reference_ability, surfaced_gaps consistency for backing: null entries)",
+    "Step 3: Enumerate abilities statically per static-enumeration.md; record 7 registrations and resolve each callback to file + line range",
+    "Step 4: Run the adversarial readonly-but-writes check per annotation-correctness.md against every callback body",
+    "Step 5: FLAG the offending read ability with FAIL, citing file + line of the $wpdb->update(...) call",
+    "Step 6: Return OK on the adversarial check for the other 5 readonly abilities and N/A for the write ability (readonly not claimed)",
+    "Step 7: Run schema lints and permission roundtrip in static-only mode",
+    "Step 8: Cross-reference error codes (non-vocabulary codes WARN, not FAIL)",
+    "Step 9: Write a structured markdown report at the user-specified path with Last updated header and Status: FAIL"
+  ],
+  "success_criteria": [
+    "Routes to wp-abilities-verify skill",
+    "Output is a markdown file at the user-specified path",
+    "Top-line status is FAIL",
+    "Annotation correctness table has exactly one FAIL row, for the ability that calls $wpdb->update, with file + line evidence",
+    "The other 5 read abilities each have a passing readonly row",
+    "The write ability has no FAIL in the annotation-correctness table",
+    "Audit doc validation section confirms well-formedness or flags specific mismatches",
+    "Report header notes Static Mode so reader knows runtime checks weren't executed"
+  ]
+}
--- a/eval/scenarios/wp-abilities-verify.json
+++ b/eval/scenarios/wp-abilities-verify.json
@@ -8,7 +8,7 @@
     "Step 3: Enumerate abilities statically per static-enumeration.md; record 7 registrations and resolve each callback to file + line range",
     "Step 4: Run the adversarial readonly-but-writes check per annotation-correctness.md against every callback body",
     "Step 5: FLAG the offending read ability with FAIL, citing file + line of the $wpdb->update(...) call",
-    "Step 6: Return OK on the adversarial check for the other 5 readonly abilities and N/A for the write ability (readonly not claimed)",
+    "Step 6: Return OK on the adversarial check for the other 5 readonly abilities and OK (write path confirmed) for the write ability (readonly=false annotation acknowledged; check only applies when readonly=true)",
     "Step 7: Run schema lints and permission roundtrip in static-only mode",
     "Step 8: Cross-reference error codes (non-vocabulary codes WARN, not FAIL)",
     "Step 9: Write a structured markdown report at the user-specified path with Last updated header and Status: FAIL"
--- a/skills/wp-abilities-api/references/delegate-helper-pattern.md
+++ b/skills/wp-abilities-api/references/delegate-helper-pattern.md
@@ -1,0 +1,215 @@
+# Delegate helper pattern
+
+When an ability's execute callback needs to hand work to an existing REST controller instead of duplicating business logic, extract a `delegate_to_rest_controller` helper. This reference documents the canonical signature, the three guards every implementation needs, the dual response-shape unwrap, and when NOT to use the helper at all.
+
+Read `plugin-family-patterns.md` first — the helper's exact shape depends on whether your plugin follows the shared-API-client pattern or the zero-arg-controllers pattern.
+
+## Why this helper exists
+
+List-style read abilities typically repeat the same four steps in every execute callback:
+
+1. Validate the plugin is initialized (class exists + dependencies resolved).
+2. Build a `WP_REST_Request` and copy the ability's `$input` onto it as params.
+3. Instantiate the backing controller and invoke the named method.
+4. Unwrap the response (controllers return either a raw array or a `WP_REST_Response`).
+
+Four abilities × this repetition = the boilerplate dominates the callback body. Extracting a helper keeps each execute callback a 3–4 line function that reads like pseudocode.
+
+## When to extract
+
+After the **second** list-style ability lands. Before the third ability gets added, refactor the duplicated code out of the first two execute callbacks into a private static helper. Convert subsequent abilities to call the helper as you add them.
+
+If the plugin only ever ships one or two abilities, inline the code. The helper's payoff is at 3+ callers.
+
+## Canonical signature (Pattern A — shared-API-client)
+
+```php
+/**
+ * Delegate an ability's execute callback to a REST controller.
+ *
+ * Builds a WP_REST_Request from the ability's input, instantiates the
+ * backing controller with the plugin's shared API client, invokes the
+ * named method, and normalizes the return so callers always see
+ * array|\WP_Error. Controllers in this plugin return either a
+ * WP_REST_Response or a raw array; the helper unwraps both shapes.
+ *
+ * Not used by zero-arg abilities — those call their controller directly
+ * so we don't synthesize a WP_REST_Request just to discard it.
+ *
+ * @param string     $controller_class Fully-qualified controller class (no leading backslash).
+ * @param string     $method           Controller method to invoke on the built request.
+ * @param string     $route            REST route string used when constructing WP_REST_Request.
+ * @param array|null $input            Ability input; each key/value becomes a request param.
+ * @param string     $http_method      HTTP method for WP_REST_Request. Defaults to 'GET'; writes pass 'POST'.
+ * @return array|\WP_Error             Response payload as an array, or WP_Error on failure.
+ */
+private static function delegate_to_rest_controller(
+    string $controller_class,
+    string $method,
+    string $route,
+    ?array $input,
+    string $http_method = 'GET'
+) {
+    $fqcn = '\\' . $controller_class;
+
+    // Guard 1: controller class is loaded.
+    if ( ! class_exists( $fqcn ) ) {
+        return new \WP_Error(
+            '<plugin>_not_initialized',
+            __( '<Plugin> is not initialized.', '<text-domain>' )
+        );
+    }
+
+    // Guard 2: shared API client is available.
+    $api_client = null;
+    if ( class_exists( '\<Plugin_Main_Class>' ) && method_exists( '\<Plugin_Main_Class>', 'get_<client_accessor>' ) ) {
+        $api_client = \<Plugin_Main_Class>::get_<client_accessor>();
+    }
+    if ( null === $api_client ) {
+        return new \WP_Error(
+            '<plugin>_not_initialized',
+            __( '<Plugin> is not initialized.', '<text-domain>' )
+        );
+    }
+
+    // Build the request.
+    $request = new \WP_REST_Request( $http_method, $route );
+    if ( null !== $input ) {
+        foreach ( $input as $param => $value ) {
+            $request->set_param( $param, $value );
+        }
+    }
+
+    // Invoke + guard 3: WP_Error short-circuit + dual-shape unwrap.
+    $controller = new $fqcn( $api_client );
+    $response   = $controller->{$method}( $request );
+
+    if ( is_wp_error( $response ) ) {
+        return $response;
+    }
+    if ( $response instanceof \WP_REST_Response ) {
+        $data = $response->get_data();
+        return is_array( $data ) ? $data : [];
+    }
+    return is_array( $response ) ? $response : [];
+}
+```
+
+Positional (not named) arguments on purpose — the helper is called from every list ability and keyword-args for PHP 8 would add visual noise. Order is intentional: controller, method, route, input, http_method (the defaultable tail).
+
+## The three guards
+
+### 1. `class_exists` on the controller
+
+Execute callbacks can be reached on sites where the plugin's classes haven't loaded (tests, WP-CLI with limited bootstrap, admin pages with conditional autoloading). Check is cheap; without it a missing class produces a fatal.
+
+Note `'\\' . $controller_class` — accept controller class names without a leading backslash (readable in config arrays) and re-add it so `class_exists` and `new $fqcn(...)` both root-resolve.
+
+### 2. Required dependency (API client) null-check
+
+For Pattern A plugins the accessor is a `public static` method on the main plugin class. Both `class_exists` on the plugin class AND `method_exists` on the accessor are guarded because either could be absent during partial bootstraps. A null return from the accessor is also treated as "not initialized".
+
+Missing this argument produces `ArgumentCountError: Too few arguments to function <Controller>::__construct(), 0 passed` — a PHP fatal. The standardized `<plugin>_not_initialized` error code is documented in `error-code-vocabulary.md`.
+
+For Pattern B plugins, this guard is either skipped entirely (zero-arg constructor) or replaced with construction-time scalar args passed in via an optional `constructor_args` parameter.
+
+### 3. `is_wp_error` short-circuit
+
+Before trying to unwrap the response, check if it's already a `WP_Error`. Several controllers return `WP_Error` for both validation failures and upstream failures; that error needs to bubble up intact so the agent can see the upstream code.
+
+## Dual response-shape unwrap
+
+```php
+if ( $response instanceof \WP_REST_Response ) {
+    $data = $response->get_data();
+    return is_array( $data ) ? $data : [];
+}
+return is_array( $response ) ? $response : [];
+```
+
+Real WordPress REST controllers return either:
+- A `WP_REST_Response` (the output of `rest_ensure_response( ... )` or `new WP_REST_Response( ... )`), or
+- A raw array (typical when a controller delegates to an internal request-handler class that returns the decoded transport payload directly).
+
+Both happen. The helper unwraps `WP_REST_Response` via `get_data()` and passes raw arrays through. The `is_array(...) ? ... : []` coalesce defends against a non-array return type that the ability's `output_schema` would have rejected downstream anyway — returning `[]` fails safely rather than leaking a surprising type.
+
+## When NOT to use the helper
+
+Two categories of abilities bypass the helper and call the backing directly:
+
+### Zero-arg backing methods
+
+If the backing method takes no `WP_REST_Request` parameter, don't synthesize a request just to discard it. Call the method directly:
+
+```php
+public static function execute_get_overview( $input = null ) {
+    if ( ! class_exists( '\My_Plugin_REST_Overview_Controller' ) ) {
+        return new \WP_Error( '<plugin>_not_initialized', __( '<Plugin> is not initialized.', '<text-domain>' ) );
+    }
+
+    $api_client = /* ... same accessor check ... */;
+    if ( null === $api_client ) {
+        return new \WP_Error( '<plugin>_not_initialized', __( '<Plugin> is not initialized.', '<text-domain>' ) );
+    }
+
+    $controller = new \My_Plugin_REST_Overview_Controller( $api_client );
+    $response   = $controller->get_overview(); // No $request argument.
+
+    if ( is_wp_error( $response ) ) {
+        return $response;
+    }
+    return is_array( $response ) ? $response : [];
+}
+```
+
+### Abilities that use a non-REST service
+
+If the execute callback reaches a service directly (e.g. `My_Plugin::get_account_service()->get_cached_account_data()`) rather than a REST controller, stay on the direct path. The ability is not REST-backed and the helper would add a construction step for no gain.
+
+## Rule of thumb
+
+**If the backing method's signature includes `WP_REST_Request`, use the helper. Otherwise direct-path.**
+
+## Conversion pattern — before and after
+
+Before extraction (inline in the execute callback):
+
+```php
+public static function execute_get_things( $input = null ) {
+    if ( ! class_exists( '\My_Plugin_REST_Things_Controller' ) ) {
+        return new \WP_Error( '<plugin>_not_initialized', __( '<Plugin> is not initialized.', '<text-domain>' ) );
+    }
+    $api_client = \My_Plugin::get_api_client();
+    if ( ! $api_client ) {
+        return new \WP_Error( '<plugin>_not_initialized', __( '<Plugin> is not initialized.', '<text-domain>' ) );
+    }
+    $request = new \WP_REST_Request( 'GET', '/my-plugin/v1/things' );
+    foreach ( (array) $input as $param => $value ) {
+        $request->set_param( $param, $value );
+    }
+    $controller = new \My_Plugin_REST_Things_Controller( $api_client );
+    $response   = $controller->get_things( $request );
+    // ... unwrap ...
+}
+```
+
+After extraction:
+
+```php
+public static function execute_get_things( $input = null ) {
+    return self::delegate_to_rest_controller(
+        'My_Plugin_REST_Things_Controller',
+        'get_things',
+        '/my-plugin/v1/things',
+        is_array( $input ) ? $input : null
+    );
+}
+```
+
+Note the `is_array( $input ) ? $input : null` — the helper signature is `?array`, and passing `null` tells it to build the request with no params at all. This matters because the Abilities API sometimes invokes a callback with no input object.
+
+## Related references
+
+- `plugin-family-patterns.md` — choosing the helper's constructor shape.
+- `error-code-vocabulary.md` — the `<plugin>_not_initialized` code and its siblings.
+- `input-schema-gotchas.md` — callbacks for list abilities often need `per_page` pagination translation BEFORE calling the helper.
--- a/skills/wp-abilities-api/references/delegate-helper-pattern.md
+++ b/skills/wp-abilities-api/references/delegate-helper-pattern.md
@@ -1,8 +1,8 @@
 # Delegate helper pattern
 
-Once you've decided delegation is the right shape for an ability — that is, the backing REST controller is a pure data-fetch, the operation is a read, and the ability runs predominantly inside REST contexts (see `shared-core-service.md` for when the answer is "no, extract a shared service" instead) — extract a `delegate_to_rest_controller` helper rather than open-coding the request-build/dispatch/unwrap pipeline in each execute callback. This reference documents the canonical signature, the three guards every implementation needs, the dual response-shape unwrap, and when NOT to use the helper at all.
+Once you've decided delegation is the right shape for an ability — that is, the backing REST controller is a pure data-fetch, the operation is a read, and the ability runs predominantly inside REST contexts (see `shared-core-service.md` for when the answer is "no, extract a shared service" instead) — extract a `delegate_to_rest_controller` helper rather than open-coding the request-build/dispatch/unwrap pipeline in each execute callback. This reference documents one helper shape that works, the three guards every implementation needs, the dual response-shape unwrap, and when NOT to use the helper at all. Adapt the exact signature to the plugin you're working in — the guards and unwrap matter more than the parameter list.
 
-Read `plugin-family-patterns.md` first — the helper's exact shape depends on whether your plugin follows the shared-API-client pattern or the zero-arg-controllers pattern.
+Read `plugin-family-patterns.md` first — the helper's exact shape depends on how the backing controller is constructed (shared API client vs. zero-arg).
 
 ## Why this helper exists
 
@@ -21,7 +21,7 @@ After the **second** list-style ability lands. Before the third ability gets add
 
 If the plugin only ever ships one or two abilities, inline the code. The helper's payoff is at 3+ callers.
 
-## Canonical signature (Pattern A — shared-API-client)
+## Helper shape
 
 ```php
 /**
@@ -107,11 +107,11 @@ Note `'\\' . $controller_class` — accept controller class names without a lead
 
 ### 2. Required dependency (API client) null-check
 
-For Pattern A plugins the accessor is a `public static` method on the main plugin class. Both `class_exists` on the plugin class AND `method_exists` on the accessor are guarded because either could be absent during partial bootstraps. A null return from the accessor is also treated as "not initialized".
+When the controller takes a shared API client as a constructor argument, the accessor is typically a `public static` method on the main plugin class. Guard both `class_exists` on the plugin class AND `method_exists` on the accessor because either could be absent during partial bootstraps. Treat a null return from the accessor as "not initialized".
 
 Missing this argument produces `ArgumentCountError: Too few arguments to function <Controller>::__construct(), 0 passed` — a PHP fatal. The standardized `<plugin>_not_initialized` error code is documented in `error-code-vocabulary.md`.
 
-For Pattern B plugins, this guard is either skipped entirely (zero-arg constructor) or replaced with construction-time scalar args passed in via an optional `constructor_args` parameter.
+When the controller takes no constructor args, skip this guard entirely. When it takes only simple scalars (e.g. a post-type string), pass them in via an optional `constructor_args` parameter on the helper.
 
 ### 3. `is_wp_error` short-circuit
 
--- a/skills/wp-abilities-api/references/delegate-helper-pattern.md
+++ b/skills/wp-abilities-api/references/delegate-helper-pattern.md
@@ -1,6 +1,6 @@
 # Delegate helper pattern
 
-When an ability's execute callback needs to hand work to an existing REST controller instead of duplicating business logic, extract a `delegate_to_rest_controller` helper. This reference documents the canonical signature, the three guards every implementation needs, the dual response-shape unwrap, and when NOT to use the helper at all.
+Once you've decided delegation is the right shape for an ability — that is, the backing REST controller is a pure data-fetch, the operation is a read, and the ability runs predominantly inside REST contexts (see `shared-core-service.md` for when the answer is "no, extract a shared service" instead) — extract a `delegate_to_rest_controller` helper rather than open-coding the request-build/dispatch/unwrap pipeline in each execute callback. This reference documents the canonical signature, the three guards every implementation needs, the dual response-shape unwrap, and when NOT to use the helper at all.
 
 Read `plugin-family-patterns.md` first — the helper's exact shape depends on whether your plugin follows the shared-API-client pattern or the zero-arg-controllers pattern.
 
--- a/skills/wp-abilities-api/references/delegate-helper-pattern.md
+++ b/skills/wp-abilities-api/references/delegate-helper-pattern.md
@@ -135,7 +135,33 @@ Both happen. The helper unwraps `WP_REST_Response` via `get_data()` and passes r
 
 ## When NOT to use the helper
 
-Two categories of abilities bypass the helper and call the backing directly:
+Three categories of abilities bypass the helper and call the backing directly:
+
+### Backing request class can be invoked without the controller
+
+If the REST controller's only role is to translate `WP_REST_Request` into a call to an underlying request class (e.g., `Things_Request::from_rest_request( $request )->execute()`) and adds no validation, permission logic, or orchestration on top, bypass the controller and call the request class directly:
+
+```php
+public static function execute_get_things( $input = null ) {
+    if ( ! class_exists( '\My_Plugin\Things_Request' ) ) {
+        return new \WP_Error( '<plugin>_not_initialized', __( '<Plugin> is not initialized.', '<text-domain>' ) );
+    }
+
+    $request = new \WP_REST_Request( 'GET', '/my-plugin/v1/things' );
+    foreach ( (array) $input as $param => $value ) {
+        $request->set_param( $param, $value );
+    }
+
+    $things_request = \My_Plugin\Things_Request::from_rest_request( $request );
+    $rows           = $things_request->execute();
+
+    return is_array( $rows ) ? $rows : [];
+}
+```
+
+The `WP_REST_Request` object exists only to satisfy `from_rest_request()`'s parameter contract — nothing dispatches it. This is the shortest path through the layers: no controller construction, no controller-emitted side effects, and (when the request class is hit before any REST traffic in the lifecycle) no `rest_api_init` cost. See `shared-core-service.md` for the broader discussion of REST-path side effects and the first-call bootstrap cost on cold paths.
+
+The tradeoff is real: bypassing the controller bypasses anything the controller does. If the controller runs validation that the request class doesn't, or emits an audit hook the request class doesn't, that work is lost on the ability path. Use this shape when the controller is genuinely a thin wrapper, not when it's doing work you'd want to keep.
 
 ### Zero-arg backing methods
 
@@ -168,7 +194,7 @@ If the execute callback reaches a service directly (e.g. `My_Plugin::get_account
 
 ## Rule of thumb
 
-**If the backing method's signature includes `WP_REST_Request`, use the helper. Otherwise direct-path.**
+**If the backing method takes `WP_REST_Request` and the controller adds something beyond delegating to an underlying request class (validation, permissions, orchestration, side effects you want to keep), use the helper. If the controller is a thin wrapper over a request class, call the request class directly. Otherwise direct-path.**
 
 ## Conversion pattern — before and after
 
--- a/skills/wp-abilities-api/references/delegate-helper-pattern.md
+++ b/skills/wp-abilities-api/references/delegate-helper-pattern.md
@@ -44,11 +44,11 @@ If the plugin only ever ships one or two abilities, inline the code. The helper'
  * @return array|\WP_Error             Response payload as an array, or WP_Error on failure.
  */
 private static function delegate_to_rest_controller(
-    string $controller_class,
-    string $method,
-    string $route,
-    ?array $input,
-    string $http_method = 'GET'
+    $controller_class,
+    $method,
+    $route,
+    $input,
+    $http_method = 'GET'
 ) {
     $fqcn = '\\' . $controller_class;
 
--- a/skills/wp-abilities-api/references/input-schema-gotchas.md
+++ b/skills/wp-abilities-api/references/input-schema-gotchas.md
@@ -1,0 +1,221 @@
+# Input schema gotchas
+
+Three surprises the Abilities API's `input_schema` will ship with if you rely on schema declarations alone. All three have been caught in real plugin work after the schema looked correct and tests passed; each has a defensive pattern that makes the execute callback robust regardless.
+
+## 1. Schema `default` values are NOT injected into execute callback input
+
+### Problem
+
+`wp_register_ability()` lets you declare per-property defaults in `input_schema`:
+
+```php
+'input_schema' => [
+    'type'       => 'object',
+    'properties' => [
+        'submit' => [
+            'type'        => 'boolean',
+            'default'     => false,
+            'description' => __( 'Whether to submit evidence for review.', '<text-domain>' ),
+        ],
+    ],
+],
+```
+
+The Abilities API's validate path enforces `type` and `required`, but it does NOT populate missing properties with their declared `default`. If an agent invokes the ability with `{ dispute_id: "dp_..." }` and omits `submit`, the execute callback receives `$input` **without** a `submit` key — it does not get `$input['submit'] = false`.
+
+### Symptoms
+
+- Boolean defaults silently become "undefined" in the callback and any `if ( $input['submit'] )` check compares against `null`, which works by accident but fails `isset` checks and strict-type branches.
+- Integer pagination defaults like `'page' => [ 'default' => 1 ]` never reach the backing controller, so pagination falls back to the controller's internal default rather than the one declared in the schema.
+- Object defaults (`'metadata' => [ 'default' => [] ]`) become `null` rather than `[]` and any `foreach ( $input['metadata'] as ... )` hits a type error.
+
+### Fix — normalize defaults explicitly in the execute callback
+
+```php
+public static function execute_submit_evidence( $input = null ) {
+    if ( ! is_array( $input ) ) {
+        $input = [];
+    }
+
+    // Apply schema defaults the Abilities API does not inject.
+    if ( ! array_key_exists( 'submit', $input ) || null === $input['submit'] ) {
+        $input['submit'] = false;
+    }
+    $input['submit'] = (bool) $input['submit'];
+
+    if ( ! array_key_exists( 'metadata', $input ) || null === $input['metadata'] ) {
+        $input['metadata'] = [];
+    }
+
+    // ... rest of callback
+}
+```
+
+The `array_key_exists` + null-check pattern catches both "missing key" and "explicit null" (some serializers produce nulls for optional fields).
+
+Keep the declared `default` in `input_schema` anyway — it documents the expected behavior for anyone reading the registration and is visible to agents in the schema introspection endpoints. Just don't rely on it for runtime population.
+
+## 2. Pagination parameter-name drift
+
+### Problem
+
+The `input_schema` convention across most WordPress REST endpoints is `per_page` (matching WP core REST list endpoints and `WP_REST_Controller`'s `get_collection_params()`). Some plugin REST controllers, however, delegate to an internal request-builder class that reads a different key (e.g. `pagesize` or `page_size`), typically for historical reasons.
+
+If the ability's `input_schema` exposes `per_page` and the backing controller reads `pagesize`, the agent's `per_page: 50` silently never reaches pagination — the value is accepted, forwarded verbatim to the backing, and then ignored. Agents keep getting the default page size and suspect their filter is broken.
+
+### Symptoms
+
+- List abilities return the backing controller's default page size regardless of the agent's `per_page` input.
+- Integration tests that only check "a list came back" pass; only a test asserting `count( $response['data'] )` catches it.
+- Harness runs show the raw response count matching the default (25, 10, etc.) for every call.
+
+### Detection heuristic
+
+Before shipping a paginated ability, grep the backing controller's call chain:
+
+```bash
+# Check the backing controller and anything it delegates to.
+grep -rn "pagesize\|page_size" <path/to/backing-controller.php> <path/to/request-helpers/>
+```
+
+If `pagesize` (or any non-`per_page` key) appears in the chain and the ability's schema exposes `per_page`, the execute callback MUST translate.
+
+### Fix — translate before delegating
+
+```php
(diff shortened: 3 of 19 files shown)


skills/wp-abilities-verify/references/annotation-correctness.md
```diff
@@ -0,0 +1,304 @@
...
+#    a) [ ClassName::class, 'method' ] — look in ClassName for `public function method` or `public static function method`.
+#    b) [ $this, 'method' ] — look in the current class.
+#    c) 'function_name' — look for `function function_name` top-level.
+#    d) An anonymous function — the body is literally there.
+```
+
+Record the file + starting line + ending line of the callback. All
+subsequent greps target that byte range.
+
+### Step 2 — readonly-but-writes check
```
nerrad: Should the skill be so deterministic here? I'm wary that it locks behaviour to current model behaviour and prevents potentially better approaches with newer models that are more intelligent.

skills/wp-abilities-verify/references/annotation-correctness.md
```diff
@@ -0,0 +1,304 @@
...
+
+# Inline non-GET WP_REST_Request construction (without the delegate helper).
+grep -nE "new\s+\\\\?WP_REST_Request\s*\(\s*['\"](POST|PUT|DELETE|PATCH)" <file>
+
+# Filesystem writes — durable state changes that tools/list cannot describe.
+grep -nE '\b(file_put_contents|wp_upload_bits|fwrite|fputs|unlink|rename|mkdir|rmdir|copy|move_uploaded_file)\s*\(' <file>
+grep -nE 'WP_Filesystem(_Direct)?\s*->\s*(put_contents|delete|move|copy|mkdir|rmdir)' <file>
+
+# Cron and scheduler writes — mutate the cron options table even on "readonly" abilities.
+grep -nE '\bwp_(schedule_event|schedule_single_event|reschedule_event|unschedule_event|clear_scheduled_hook)\s*\(' <file>
```
nerrad: It appears the method regex doesn't catch direct mutators like `->save()`, `->delete()`, `->add_order_note` or `->set_status()`. If you are going to keep deterministic examples in the skill (which I don't think we should do, because it specifically keeps the model from intelligently deriving those - and this example already demonstrates missing them), maybe include things like:

- include method calls where the verb is followed by `(` or `_`
- include additive write verbs such as `add`, `set`, `save`, `delete`, `trash`.

**My recommendation is to remove these specific specifiers. You can provide a few examples of verbs, but the model should be encouraged to view those _as_ examples, not deterministic instructions**

skills/wp-abilities-verify/references/schema-lints.md
```diff
@@ -0,0 +1,239 @@
...
+```
+
+**Why:** a default that evaluates per-call is both non-deterministic
+(twin invocations can differ, violating `idempotent: true`) and
+surprising to agents that expect defaults to be static values.
+
+**Check:** each `'default'` value must be a scalar literal (`true`,
+`false`, integer, float, quoted string, `null`) or an empty array
+literal. A function call, variable reference, or computed expression →
+FAIL.
```
nerrad: This says defaults must be primitive, null, or an empty array. But `input-schema-gotchas.md` correctly says zero-arg object ab ilities need a top-level schema default like `(object) array()` because `normalize_input()` applies top-level defaults. See https://github.com/rtio/agent-skills/blob/bb1de00133d64698b78114c9712749ac814c027c/skills/wp-abilities-api/references/input-schema-gotchas.md#L205-L222 and [`WP_Ability::normalize_input()`](https://developer.wordpress.org/reference/classes/wp_ability/normalize_input/). As written, it looks like `verify` could end up flagging the recommended fix here as invalid.

skills/wp-abilities-verify/references/static-enumeration.md
```diff
@@ -0,0 +1,262 @@
...
+## Step 2 — extract each ability name
+
+The first argument to `wp_register_ability` is the ability name. Usually
+a literal string:
+
+```bash
+# Greedy capture of single- or double-quoted first argument.
+grep -rn --include='*.php' -E "wp_register_ability\s*\(\s*['\"]([^'\"]+)['\"]" <plugin-root>/ \
+    | sed -nE "s/.*wp_register_ability\s*\(\s*['\"]([^'\"]+)['\"].*/\1/p"
+```
```
nerrad, line 56: As metioned elsewhere, I'm wary that the deterministic instructions in the skill risks model indirection. This specific case seems brittle because it assumes `wp_register_ability(` and the first string argument appear on the same line, when that's not always the case (and even the example given in this file is not on the same line...).

skills/wp-abilities-verify/references/annotation-correctness.md
```diff
@@ -0,0 +1,304 @@
...
+The same logic applies to `destructive: false` ("won't delete anything")
+and `idempotent: true` ("same input, same result, same effect").
+
+## The three claims
+
+| Annotation | What it promises | What to check |
+|---|---|---|
+| `readonly: true` | No writes. GET-style side-effect-free. | No `wpdb->insert/update/delete`, `update_option`, `wp_insert_*`, non-GET delegates, etc. |
+| `destructive: false` | Won't irreversibly destroy data or forfeit money. | No `wp_delete_*`, `wp_trash_post`, `->refund`, `->cancel`, `->close_dispute`. |
+| `idempotent: true` | Same input always produces same effect. | No `rand()`, nonce generation, sequence increments, timestamps-in-response. Runtime: twin invocations byte-identical. |
```
nerrad: [Core describes idempotency](https://github.com/WordPress/wordpress-develop/blob/a5905ab11bedac9ba835b444330e74d996bf597b/src/wp-includes/abilities-api/class-wp-ability.php#L47-L48) as "no additional effect on the environment". So its more about repeated calls having no additional environmental effect, not necessarily byte-identical return values.

Although I know a stricter reading of `idempotent` does have the meaning outlined here, but don't think we should have the skill imply more than what the source implies. 

I'm also wary that the length instruction set here risks becoming stale with future changes to the Abilities API and how the annotations are used which will be an indirection for any agents. For example, The Abilities REST `run` controller [operationalizes these values into allowed HTTP methods](https://github.com/WordPress/wordpress-develop/blob/a5905ab11bedac9ba835b444330e74d996bf597b/src/wp-includes/rest-api/endpoints/class-wp-rest-abilities-v1-run-controller.php#L110-L116).

nerrad (changes requested):
As I reviewed the current state of the `wp-abilities-verify` skill, I'm a bit wary that it might be too exhaustive/wordy and risks agent indirection or getting stale with the source code. Especially when smarter models come along (this has already happened with some of the attempts I've made with using this). So comments will reflect that.

Another minor thing to note is that the the repo's README's "Available Skills" table doesn't list `wp-abilities-audit` or `wp-abilities-verify`.

skills/wp-abilities-verify/references/annotation-correctness.md
rtio: Logging this as the cross-cutting thread that runs through the whole review — the references want to flex from prescriptive grep recipes toward principles + a few examples + agent-derived patterns. Holding it for a focused follow-up pass rather than reshaping in this round; landed the bug-fix subset first.

The five commits in this push (35f60d8, 16eb6a1, ee85b28, c5f962a, d8ea032) cover the specific items each of the other threads called out. Want me to open a follow-up issue to track the broader reshape so it does not get lost?

skills/wp-abilities-verify/references/annotation-correctness.md
rtio: Fixed in 16eb6a1.

The trailing `_` was filtering out direct calls — `verb(` never matched. New pattern is `verb(_|\()` so both verb-prefixed names (`->save_record()`) and direct calls (`->save()`) match. Added `add`, `set`, and `trash` to the verb list per your suggestion. Verbatim regex now:

```
->(add|set|save|create|insert|update|delete|remove|destroy|purge|trash|archive|restore|store|write|disable|enable|activate|deactivate|revoke|grant)(_|\()
```

The verb list is now explicitly framed as a representative starting set, not exhaustive — a partial step toward your broader recommendation. The full "remove these specific specifiers" reshape is on the deferred follow-up track logged on the philosophical thread.

skills/wp-abilities-verify/references/schema-lints.md
rtio: Good catch — fixed in c5f962a.

Lint 5 now accepts `(object) array()`, `(object) []`, `new stdClass()`, and all-literal arrays as valid defaults. Cross-linked `input-schema-gotchas.md` §4 explicitly in the lint copy so the agreement between the two skills is visible to anyone reading either one. Function-call defaults like `gmdate('c')` or `wp_generate_uuid4()` still FAIL because the value varies per call — the lint is now about "evaluates to the same shape every call," which the cast and `new stdClass()` do.

skills/wp-abilities-verify/references/static-enumeration.md
rtio: Fixed in ee85b28.

You are right — the file's own canonical example is multi-line, and the regex did not match it. Step 2 now recommends `rg --multiline --pcre2` as the primary path with `pcregrep -M` and `perl -0777` documented as fallbacks. Single-line grep stays as a fast inline-form check but is no longer the only option offered.

The broader "trust the agent more, prescribe less" thread lands on the deferred reshape track.

skills/wp-abilities-verify/references/annotation-correctness.md
rtio: Fixed in 35f60d8.

The three-claims table now matches core's docblock at `class-wp-ability.php:47-48` ("no additional effect on its environment"). Added a cross-reference to the run controller's HTTP method routing (`class-wp-rest-abilities-v1-run-controller.php:110-116`) so the load-bearing semantic is grounded in core, not in a stricter reading.

The static check stops flagging `rand()`, `wp_create_nonce()`, and `time()` — they don't violate environmental idempotency. The new patterns target what almost always reflects real environmental accumulation: counter writes, per-call cron schedules (`wp_schedule_event` without a "is it scheduled?" guard), append-only `$wpdb->insert` calls, and visible sequence increments.

Runtime twin-invocation is reframed as a heuristic, not a verdict: when responses differ, the reply distinguishes "embeds a timestamp / nonce / random ID → still idempotent" from "reflects a counter that grew → real environmental change → FAIL." For ambiguous cases, supplement with a state diff between the two calls.

rtio:
Hey Darren — pushed five focused fixes addressing the specific issues; replied inline on each thread with detail.

| Commit | Thread | What changed |
|---|---|---|
| `35f60d8` | idempotent definition | Match core's "no additional effect on its environment" wording (`class-wp-ability.php:47-48`); cross-ref the run controller's HTTP routing as the load-bearing semantic; static check no longer flags `rand()` / `wp_create_nonce()` / `time()`; runtime twin-invocation reframed as heuristic, not verdict |
| `16eb6a1` | mutator regex | Pattern is now `verb(_|\()` so `->save()` / `->delete()` / `->add_order_note()` / `->set_status()` all match; added `add` / `set` / `trash`; verb list explicitly framed as illustrative |
| `ee85b28` | multi-line registration | `rg --multiline --pcre2` as primary, with `pcregrep -M` / `perl -0777` fallbacks; the canonical multi-line example now matches the regex used to enumerate it |
| `c5f962a` | Lint 5 vs `(object) array()` | Lint accepts `(object) array()`, `(object) []`, `new stdClass()`, all-literal arrays; cross-linked `input-schema-gotchas.md` §4 so the two skills no longer quietly disagree |
| `d8ea032` | README | Added `wp-abilities-audit` and `wp-abilities-verify` to the Available Skills table |

On the broader thread that runs through the whole review — that the references are over-prescriptive and risk locking behavior to current model capability — I think you're right. The right move is to reshape from "exhaustive grep playbook" to "principles + a few examples + agent-derived patterns." That's a substantive change and I'd rather scope it as a separate pass once these bug fixes are in. Happy to open a follow-up issue if you'd like to track it explicitly.

rtio:
Hey @nerrad  — agreed and pivoted. The round-2 bug fixes landed first (idempotent definition, Lint 5 vs `(object) array()`, README); for the substantive part of your review I cut the prescriptive grep playbooks across the references so the skill reads as principles + a few illustrative examples + "agent operationalizes for the plugin's vocabulary."

Round-3 commits (`dad0196..f0070f6`):

| Commit | Effect |
|---|---|
| `dad0196` | Slim `annotation-correctness.md` (333 → 145), `schema-lints.md` (250 → 111), `static-enumeration.md` (280 → 126). Drops the exhaustive write-pattern playbook, the severity table, the hardcoded false-positive allow-list, and the embedded grep recipes for each schema lint. Keeps the Why framing, the three-claims table grounded in core's docblock, the HTTP method routing reference, the known blind spots, the `// verify-ignore` suppression, the runtime heuristic, and the report formats. |
| `f0070f6` | Slim `permission-roundtrip.md` (301 → 164) and tighten `SKILL.md`. The static grep block in permission-roundtrip — the most prescriptive part — is gone; the Shape A-F catalogue compresses into a single classification table. Runnable wp eval block stays as-is. |

Skill footprint went from ~1,900 lines (after round-2) to 1,249 in this push — roughly a one-third cut. The runnable harnesses (`runtime-harness.md`, the wp eval block in `permission-roundtrip.md`, `audit-schema-validation.md`) are preserved because they're concrete runbooks rather than pattern catalogues.

Eval scenario still passes (the agent finds `$wpdb->update` in a `readonly: true` callback either way; the slim doesn't break that).

PR body rewritten to reflect the lighter framing and the iteration history. Open to further trimming if you see specific sections that still pack too much; my read is the runnable harnesses pull their weight at their current size, but happy to revisit.

skills/wp-abilities-verify/references/annotation-correctness.md
rtio: Did the reshape in this PR rather than as a follow-up after your slack note. Round-3 commits `dad0196..f0070f6` slim the references from prescriptive grep playbooks to principles + a few illustrative examples + agent-derived patterns. Top-level PR comment has the per-file numbers; PR body rewritten to match.

skills/wp-abilities-verify/references/permission-roundtrip.md
```diff
@@ -0,0 +1,164 @@
...
+actually gates on a real capability — statically (by source
+inspection) and, in runtime mode, by exercising the gate against
+unauthenticated, subscriber, and admin contexts.
+
+## Background — what `permission_callback` actually receives
+
+When an ability is invoked, the registered `permission_callback` is
+called through `WP_Ability::check_permissions( $input )`, which
+dispatches to `WP_Ability::invoke_callback( $callback, $input )`
+(`class-wp-ability.php` lines 541-551 and 507-526 in WordPress core).
```
enejb: Should we be this specific? I would expect this to change and potentially create confusion down the line.

skills/wp-abilities-verify/references/runtime-harness.md
```diff
@@ -0,0 +1,360 @@
...
+- `npx wp-env start` — projects using `@wordpress/env` without a custom
+  npm script wrapper.
+- `jetpack docker up` — Jetpack-family monorepo packages.
+- `composer install && composer test-php --setup-only` — package-local
+  test bootstraps that don't run a full WordPress install.
+- `docker-compose up -d` — plugin-specific dev Docker stacks.
+
+If `AGENTS.md` doesn't document it, ask the user rather than guessing.
+Record the env-up command + the corresponding wp-cli invocation (e.g.
+`npx wp-env run cli wp`, `jetpack docker cli wp`) and use them uniformly
```
enejb: This seems to be jetpack specific.

skills/wp-abilities-verify/references/runtime-harness.md
```diff
@@ -0,0 +1,360 @@
...
+
+Read the plugin's `AGENTS.md` for the canonical env bring-up command. Do
+NOT assume `npm run wp-env start` works for every plugin. Common patterns
+seen in real plugin trees:
+
+- `npm run wp-env start` — projects using `@wordpress/env` with a checked-in
+  `.wp-env.json` (typical for many WooCommerce-family plugins).
+- `npx wp-env start` — projects using `@wordpress/env` without a custom
+  npm script wrapper.
+- `jetpack docker up` — Jetpack-family monorepo packages.
```
enejb: This seems to be jetpack specific.

skills/wp-abilities-verify/references/permission-roundtrip.md
```diff
@@ -0,0 +1,164 @@
...
+if ( ! $ability ) {
+    echo "ability not registered" . PHP_EOL;
+    exit( 1 );
+}
+
+$results = array();
+
+// Unauthenticated.
+wp_set_current_user( 0 );
+$results["anon"] = $ability->check_permissions();
```
rtio: The Background section above documents that `WP_Ability::check_permissions( $input )` receives the validated input, but every snippet here calls it with no args.

Core's signature is `public function check_permissions( $input = null )`, so this isn't a PHP fatal — the callback simply receives `null` instead of the real input. Harmless for **Shape A** callbacks (`return current_user_can('cap')`, the preferred form documented at lines 31-46) which don't read `$input` at all. But for **Shape B** callbacks that branch on `$input` (which the skill itself flags as a smell), the gate is exercised without the data a real invocation would carry, producing a misleading roundtrip result.

What if we pass `array()` (the minimal safe value for object schemas) here, or add a one-line caveat just above the snippet: *"For Shape B callbacks that read `$input`, pass a representative value (`array()` for object schemas); for Shape A this can be omitted."*

Same pattern repeats at `runtime-harness.md:207` and `runtime-harness.md:216`.

skills/wp-abilities-verify/references/schema-lints.md
```diff
@@ -0,0 +1,111 @@
...
+If an audit doc is provided and an ability has
+`reference_ability: true`, its `input_schema.required` array must be
+empty or absent. The reference ability is the smallest, safest
+bootstrap call an implementer lands first; it must work with
+`execute([])`. Required inputs on the reference ability → FAIL.
+
+(No audit provided → this lint is skipped — no reference ability is
+declared.)
+
+## Cross-reference: the four runtime gotchas
```
rtio: Two reviewers (clarity and docs-drift) independently flagged this. The heading at line 85 says **"four runtime gotchas"** and the next sentence at line 87 says **"the three runtime gotchas in input-schema-gotchas.md need defensive code."** The reconciliation (gotchas 1-3 need callback hardening; gotcha 4 is structural and motivates Lint 5's `(object) array()` default) is reconstructable three lines down, but the heading/body never name the split.

`input-schema-gotchas.md` does have four numbered gotchas (verified via `grep -n '^## '`), so "four" is the right count for the heading. `SKILL.md:138` is **internally consistent** — it says "four" and enumerates four in the parenthetical, so the issue is localized to this file.

What if we restate the heading as *"Cross-reference: gotchas 1-3 (callback hardening) and gotcha 4 (structural default)"*, or keep "four" and rephrase the body to make the 1-3 / 4 split explicit?

skills/wp-abilities-verify/references/annotation-correctness.md
```diff
@@ -0,0 +1,145 @@
...
+
+Each finding gets one row in the run's "Annotation correctness" table:
+
+```markdown
+| Ability | Claim | Result | Evidence |
+|---|---|---|---|
+| myplugin/get-things | readonly=true | OK | callback reads only |
+| myplugin/get-things-with-counts | readonly=true | FAIL | `src/Abilities/Things.php:142`: `$wpdb->update( $table, ... )` |
+| myplugin/submit-thing | destructive=false | OK | no destructive patterns |
+| myplugin/submit-thing | idempotent=false | N/A | not claimed idempotent |
```
rtio: The example row contains:

```
| myplugin/submit-thing | idempotent=false | N/A | not claimed idempotent |
```

Claim says `idempotent=false` and Evidence says "not claimed idempotent" — but `idempotent=false` *is* a claim (the ability asserts non-idempotency). An agent using this row as a format template will misstate every `idempotent=false` ability's declaration.

What if Evidence becomes: *"check only applies when idempotent=true; false annotation acknowledged"*?

skills/wp-abilities-verify/references/annotation-correctness.md
```diff
@@ -0,0 +1,145 @@
...
+Unit tests won't catch this class of bug because the mock the test
+constructs looks just like the real writer. What catches it is reading
+the execute callback body and comparing what it does against what the
+annotation says it does.
+
+## What each annotation promises
+
+| Annotation | What it promises (from core) |
+|---|---|
+| `readonly: true` | No writes. GET-style side-effect-free. |
```
rtio: The annotation table sets `readonly: true` = **"No writes. GET-style side-effect-free."** But the `verify-ignore` examples at lines 107-119 explicitly accept `set_transient` (cache write) and `last_read_at` (tracking write) as legitimate suppressions.

Agents who read the table first will FAIL these patterns; agents who read the examples first will suppress them. The skill's actual enforcement boundary is somewhere between, and never stated in prose.

This is the same root cause as the `idempotent=false` example row below and the eval scenario's Step 6 wording — see the review summary for a single-sentence fix that anchors all three.

skills/wp-abilities-api/references/delegate-helper-pattern.md
```diff
@@ -0,0 +1,241 @@
...
+ * so we don't synthesize a WP_REST_Request just to discard it.
+ *
+ * @param string     $controller_class Fully-qualified controller class (no leading backslash).
+ * @param string     $method           Controller method to invoke on the built request.
+ * @param string     $route            REST route string used when constructing WP_REST_Request.
+ * @param array|null $input            Ability input; each key/value becomes a request param.
+ * @param string     $http_method      HTTP method for WP_REST_Request. Defaults to 'GET'; writes pass 'POST'.
+ * @return array|\WP_Error             Response payload as an array, or WP_Error on failure.
+ */
+private static function delegate_to_rest_controller(
```
rtio, line 46: The canonical signature here uses typed params (`string $controller_class`, `?array $input`, etc.), but `plugin-family-patterns.md:183-195` explicitly notes that helper signatures stay untyped for compatibility optics, and Pattern A/B skeletons follow that rule. Both work on PHP 7.2+ (scalar types since 7.0, nullable since 7.1), so this is a style consistency issue rather than a compat bug — but an agent synthesising a helper from both files gets contradictory signals.

What if we move the inline types to `@param` phpDoc here, or add a short consistency note explaining the divergence is intentional?

skills/wp-abilities-verify/SKILL.md
```diff
@@ -0,0 +1,212 @@
...
+  order.
+
+## Escalation
+
+- Recurring legitimate pattern that trips the adversarial check across
+  multiple plugins → propose adding it to the suppression guidance in
+  `annotation-correctness.md`. Don't broaden the candidate-pattern
+  list speculatively.
+- Audit-schema validator rejects a legitimate audit → the canonical
+  schema in `wp-abilities-audit/references/audit-schema.md` has
```
rtio: Missing `../` prefix — this resolves to `skills/wp-abilities-verify/wp-abilities-audit/references/audit-schema.md`, which doesn't exist. Every other cross-skill ref in this file (lines 137, 144) uses `../`.

```diff
- `wp-abilities-audit/references/audit-schema.md`
+ `../wp-abilities-audit/references/audit-schema.md`
```

skills/wp-abilities-verify/references/annotation-correctness.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Known blind spots
+
+Static reading + grep can't reach every write. A static-mode PASS
+means "no obvious-shape violations," not "verified write-free."
+
+| Blind spot | Why static misses it | Mitigation |
+|---|---|---|
+| Indirected service writes — `$repo->persist()`, `$service->commit()`, custom verbs. | Any finite verb list drifts; domain vocabulary varies. | Inspect callbacks that touch custom services or repositories. |
+| `do_action()` whose listeners write. | Provenance ambiguity: ability looks clean; system mutates state in a listener. | Audit listeners on the action. If any writes, downgrade or split. |
```
rtio, line 104: Enhancement, not a defect — but the blind-spot table's `do_action()` row covers explicit listener writes only. Two patterns common in the WC/Jetpack family (which is likely the primary user of this skill) aren't covered:

1. **Implicit core hooks fired by WP API calls** — `wp_insert_post()`, `update_post_meta()`, `wp_create_user()` fire `save_post`/`updated_option`/etc. automatically. The WP API call IS the write; it also fires hooks as a side-effect. Agents shouldn't expect to grep for `do_action()` to catch this.
2. **Action Scheduler deferred writes** — `as_schedule_single_action()`, `WC()->queue()->schedule_single()` persist to AS tables. The callback returns cleanly with no immediately visible DB mutation, so a static grep for `$wpdb->insert` won't catch the durable side effect.

What if we add two rows to the blind-spot table — one per category — with concrete function names?

skills/wp-abilities-api/references/input-schema-gotchas.md
```diff
@@ -0,0 +1,265 @@
...
+3. `'' === $input[ ... ]` — non-empty in the strict sense. Rejects empty string only; accepts `"0"` as a legal value.
+
+Use the standardized `<plugin>_missing_<field>` error code (see `error-code-vocabulary.md`) for this case.
+
+### Add a regression-guard unit test
+
+The guard is load-bearing enough that it deserves an explicit test — otherwise someone will simplify it back to `empty()` in a future refactor:
+
+```php
+public function test_execute_returns_wp_error_when_id_not_a_string(): void {
```
rtio: **Low** — `: void` is valid PHP 7.1+ so not a compat bug, but every production code example in the skill set deliberately omits return types per `plugin-family-patterns.md`. An agent inferring the codebase's PHP style from these files sees mixed signals.

Drop the `: void` or add an inline `// void: PHP 7.1+, fine on PHP 7.2.24` comment.

skills/wp-abilities-verify/references/permission-roundtrip.md
```diff
@@ -0,0 +1,164 @@
...
+$ability = wp_get_ability( "<plugin>/<ability-name>" );
+if ( ! $ability ) {
+    echo "ability not registered" . PHP_EOL;
+    exit( 1 );
+}
+
+$results = array();
+
+// Unauthenticated.
+wp_set_current_user( 0 );
```
rtio, line 75: **Low** — repeated harness runs accumulate `verify_sub_<timestamp>` users on shared dev environments. Same pattern in `runtime-harness.md` around line 211.

What if we add `wp_delete_user( $sub_id );` at the end of the eval block, or document a periodic cleanup command?

rtio (commented):
## Review summary — verdict: COMMENT

Skill content is well-structured and the adversarial framing (catching `readonly: true` abilities that secretly write) is a real and useful niche unit tests miss. The audit↔verify boundary is clean, `verify-ignore` is precisely scoped, and the WordPress core references largely check out against `wordpress-develop` trunk.

Findings below are real but none are merge-blocking for a markdown-only skills PR. Worth a follow-up commit before this skill sees heavy use.

### Highest-leverage edit (do this one even if you do nothing else)

The `verify-ignore` boundary is defined by examples, never by a sentence. Three findings (the `readonly: true` table cell, the `idempotent=false` example row, and the eval scenario's Step 6) trace to this single gap. Adding one declarative line to `annotation-correctness.md` resolves all three:

> `readonly: true` prohibits durable writes to user/business state. Read-through cache writes (e.g. `set_transient`) and observability timestamps (e.g. `last_read_at`) are acceptable when explicitly annotated with `verify-ignore`.

### Severity recap

- **MEDIUM (7):** harness `check_permissions()` no-arg call, four-vs-three gotchas heading, `idempotent=false` example row, `readonly: true` definition vs suppression boundary, typed-vs-untyped helper signature inconsistency, missing `../` in cross-skill path, hooks-as-writes blind-spot gaps.
- **LOW (4):** `: void` style inconsistency, `verify_sub_*` user cleanup, eval scenario Step 6 wording, repo-ambiguous line citations.

### Reviewers

Run in parallel: code-clarity, docs-drift, reference-integrity, wp-architecture. Decision-critic pass downgraded two HIGHs to MEDIUM after verifying `WP_Ability::check_permissions( $input = null )` makes `$input` optional (no `ArgumentCountError`).

### Positives

- Idempotency semantics anchor cleanly to core's docblock ("no additional effect on the environment") and stay consistent across all references.
- Two-mode framing (static + runtime) is coherent; the static-mode disclaimer is repeated without contradiction.
- `verify-ignore` mechanism is precisely scoped — annotation-specific, mandatory inline reason, intentionally narrower than a blanket override.
- Audit↔verify boundary cleanly delineated; `audit-schema.md` is the stated single source of truth and `audit-schema-validation.md` defers to it without duplication.
- The `permission_callback` contract (validated `$input`, never `WP_REST_Request`) is accurately documented and the antipattern is correctly flagged as a hard FAIL — a real failure mode that ships in plugins copied from REST controllers.
- HTTP method routing table (readonly→GET, destructive&&idempotent→DELETE, otherwise POST) is anchored to the run controller's source with file + line range.
- Round-3 slim appears successful: no leftover prescriptive grep playbooks contradicting the principles+examples framing.

eval/scenarios/wp-abilities-verify.json
```diff
@@ -0,0 +1,26 @@
...
+  "expected_behavior": [
+    "Step 1: Route to wp-abilities-verify because the task is 'verify abilities registrations'",
+    "Step 2: Validate the audit doc against audit-schema-validation.md (required fields, at-most-one reference_ability, surfaced_gaps consistency for backing: null entries)",
+    "Step 3: Enumerate abilities statically per static-enumeration.md; record 7 registrations and resolve each callback to file + line range",
+    "Step 4: Run the adversarial readonly-but-writes check per annotation-correctness.md against every callback body",
+    "Step 5: FLAG the offending read ability with FAIL, citing file + line of the $wpdb->update(...) call",
+    "Step 6: Return OK on the adversarial check for the other 5 readonly abilities and N/A for the write ability (readonly not claimed)",
```
rtio: **Low** — "N/A for the write ability (readonly not claimed)" has the same root issue as the `idempotent=false` example row in `annotation-correctness.md:141`: `readonly: false` *is* an explicit claim. The Step 6 wording also leaves ambiguous whether expected output is N/A (check skipped) or OK (write confirmed) — an agent that emits N/A skips confirming the write actually happens, which misses the inverse lie (a write ability secretly acting readonly).

What if we specify `"OK (write path confirmed)"` for the `readonly: false` case?

The highest-leverage edit in the review summary (the single-sentence `verify-ignore` boundary definition) resolves this alongside M3 and M4.

skills/wp-abilities-verify/references/permission-roundtrip.md
rtio: **Low** — `class-wp-ability.php` lines 541-551 and 507-526 match `wordpress-develop` trunk exactly (verified). But the standalone `WordPress/abilities-api` feature plugin's `invoke_callback` is shorter (no try/catch), so `check_permissions` falls at lines 527-537 there — a ~14-line offset. "In WordPress core" is ambiguous between the two sources.

What if we add a one-liner noting *"line numbers reference `wordpress-develop` trunk; the feature plugin differs by ~14 lines — navigate by method name"*, or simply switch to method-name citations?

rtio: Hey Enej — fixed in 0b0a0af. You're right that pinning to specific line numbers in core makes the citations go stale silently; the next reader following the link is more likely to land on unrelated lines than the intended ones.

Replaced with stable symbol citations:

- `WP_Ability::check_permissions` + `WP_Ability::invoke_callback` for the dispatch plumbing (here, in `permission-roundtrip.md`).
- The `idempotent` annotation's docblock in `class-wp-ability.php` for the idempotent definition (`SKILL.md`, `annotation-correctness.md`, `runtime-harness.md`).
- `WP_REST_Abilities_V1_Run_Controller::validate_request_method()` for the HTTP method routing (`annotation-correctness.md`).

Extended to all instances of the pattern in the PR since the principle is consistent — the grounding the references were carrying (idempotent = "no additional effect on the environment" comes from core, not from us; HTTP method routing comes from the run controller, not from us) is preserved, just by the more durable handle.

PR body also updated to use symbol citations in the file table and verification section.

skills/wp-abilities-verify/references/runtime-harness.md
rtio: Hey Enej — fixed in b60de59. Dropped `jetpack docker up` from the env-up example list in `runtime-harness.md` Step 0; same example also removed from the SKILL.md "Inputs required" section (it was duplicated there as a "common patterns" hint).

Replaced with a neutral line:

> Plugin families with their own dev tooling will have their own bring-up command in `AGENTS.md`; follow it as documented.

The skill already hinges on "read AGENTS.md," so agents working in any plugin family with custom dev tooling still arrive at the right command — the example list just no longer endorses one family's tooling over others.

skills/wp-abilities-verify/references/runtime-harness.md
rtio: Hey Enej — fixed in b60de59. Dropped `jetpack docker cli wp` from the wp-cli invocation examples on this line; only `npx wp-env run cli wp` remains as the illustrative case. Same "read AGENTS.md" instruction covers plugin-specific tooling without naming one family explicitly.

skills/wp-abilities-verify/references/annotation-correctness.md
rtio: Fixed in f3c4e27 — the highest-leverage edit landed.

Added one declarative sentence anchored under the annotation table:

> `readonly: true` prohibits durable writes to user or business state. Read-through cache writes (e.g. `set_transient`) and observability timestamps (e.g. `last_read_at`) are acceptable when explicitly annotated with `verify-ignore`. The static check treats unannotated writes as FAILs; annotated ones pass with the reason recorded as evidence.

Also tightened the table cell for `readonly: true` to "No durable writes to user / business state. GET-style side-effect-free." so the table and the boundary sentence speak the same vocabulary.

This single edit cascades to resolve `annotation-correctness.md:141` (the `idempotent=false` example row, now reading "OK | check only applies when idempotent=true; false annotation acknowledged") and the eval scenario's Step 6 wording — both reframed from "N/A" to acknowledged-not-claimed, with affirmative confirmation on the write path.

skills/wp-abilities-verify/references/annotation-correctness.md
rtio: Fixed in f3c4e27 (as part of the M4 cascade).

The example row now reads:

```
| myplugin/submit-thing | idempotent=false | OK | check only applies when idempotent=true; false annotation acknowledged |
```

Result is `OK` rather than `N/A`, and the evidence column makes explicit that `idempotent=false` IS a claim — the claim is "this ability does not promise idempotency" — and that the adversarial check applies only when the claim is `idempotent=true`. The same wording aligns with the new `readonly`/`verify-ignore` boundary sentence anchored above.

eval/scenarios/wp-abilities-verify.json
rtio: Fixed in f3c4e27 (also cascaded from the M4 boundary edit).

Step 6 in the eval scenario now reads:

```
Return OK on the adversarial check for the other 5 readonly abilities and OK (write path confirmed) for the write ability (readonly=false annotation acknowledged; check only applies when readonly=true)
```

Same acknowledged-not-claimed framing as the example row and the boundary sentence. The "write path confirmed" wording also addresses the inverse-lie concern — the harness affirmatively notes the write happens, rather than skipping with `N/A`.

skills/wp-abilities-verify/references/permission-roundtrip.md
rtio: Fixed in 2b86824.

Both snippets — `permission-roundtrip.md`'s main runtime block and `runtime-harness.md` Check 5 — now pass `$input = array()` to `check_permissions()`. Added a one-paragraph note above the permission-roundtrip block explaining the rationale (Shape A callbacks don't read input, Shape B branches on it; `array()` is the minimal safe value for object-typed schemas, which is the common case) and instructing the agent to substitute a representative value of the declared root type for non-object schemas.

That way the gate is exercised with the data the real invocation carries even when the callback is Shape B (the smell the static check flags), without depending on every reader noticing the contract by reading the Background section first.

skills/wp-abilities-verify/references/schema-lints.md
rtio: Fixed in 57d3d6d.

Renamed the heading to:

> Cross-reference: gotchas 1-3 (callback hardening) and gotcha 4 (structural default)

And rewrote the body to make the 1-3 / 4 split explicit — gotchas 1-3 need defensive code in the execute callback (the runtime-handling category), gotcha 4 is what motivates Lint 5's `(object) array()` acceptance (the structural-default category). Both halves are now stated in prose, not reconstructable.

Count of `^## [0-9]` in `input-schema-gotchas.md` is four — verified by `grep -nE`. SKILL.md:138's "four" was already internally consistent and didn't need changing.

skills/wp-abilities-verify/SKILL.md
rtio: Fixed in 613fce7.

```diff
- `wp-abilities-audit/references/audit-schema.md`
+ `../wp-abilities-audit/references/audit-schema.md`
```

Confirmed the other two cross-skill refs in the same file (`SKILL.md:139` and `:146`) already use `../`, so the format is now consistent and the omission was the bug.

skills/wp-abilities-api/references/delegate-helper-pattern.md
rtio: Fixed in 1ba84da.

Harmonized toward the untyped form: dropped the inline types from the `delegate_to_rest_controller` signature in `delegate-helper-pattern.md`. The phpDoc above the function (which already carried `@param string $controller_class`, `@param array|null $input`, etc.) is now the single source of types — same shape `plugin-family-patterns.md:183-195` uses.

That way the canonical reference and the Pattern A/B skeletons speak the same style, and an agent synthesising a helper from either file emits matching signatures.

skills/wp-abilities-verify/references/annotation-correctness.md
rtio: Fixed in 6b09b72.

Added two rows to the blind-spot table:

- **Implicit core hooks fired by WP API calls** — `wp_insert_post()` fires `save_post`; `update_option()` fires `updated_option`; `wp_create_user()` fires `user_register`; etc. Mitigation: treat any WP write-API call as a write regardless of whether the callback also calls `do_action()`.
- **Action Scheduler / deferred writes** — `as_schedule_single_action()`, `WC()->queue()->schedule_single()`, custom job dispatchers. Mitigation: treat scheduler dispatches as writes; the "no additional effect on the environment" promise of `idempotent: true` is violated by accumulating queued jobs even when the immediate return value is constant.

Concrete function names in the table so the agent has something to grep for, with mitigations stated rather than implied.

skills/wp-abilities-verify/references/permission-roundtrip.md
rtio: Already addressed by 0b0a0af (pushed before this thread but referencing it for completeness).

All line-number citations across the verify-skill — including this one — are now method-name citations: `WP_Ability::check_permissions`, `WP_Ability::invoke_callback`, the `idempotent` annotation's docblock, and `WP_REST_Abilities_V1_Run_Controller::validate_request_method()`. Switching to symbol-based citations sidesteps the wordpress-develop-vs-feature-plugin offset entirely — the methods exist under the same names in both sources, so navigation works regardless of which one the reader has open.

skills/wp-abilities-api/references/input-schema-gotchas.md
rtio: Deferred to a follow-up commit alongside the other LOW-severity items (`verify_sub_*` cleanup in `permission-roundtrip.md:64`). The `: void` return is valid PHP 7.1+ so not blocking; the style harmonization just isn't on the critical path for this PR.

Will land before this skill sees heavy use, per the review summary's "worth a follow-up commit" framing.

skills/wp-abilities-verify/references/permission-roundtrip.md
rtio: Deferred to the same follow-up commit as M8 (`: void` style harmonization). Cleanup matters more in shared dev environments than fresh ones; pattern repeats at `runtime-harness.md:211`, so the follow-up will sweep both call sites and either add `wp_delete_user( $sub_id );` to the eval blocks or document a periodic cleanup command.

Will land before this skill sees heavy use, per the review summary's "worth a follow-up commit" framing.

skills/wp-abilities-api/references/input-schema-gotchas.md
rtio: Landed in b054288.

Dropped the `: void` from the test method signature so it matches the untyped style established in `plugin-family-patterns.md:183-195`. The method's no-return shape is already obvious from the assertion-only body; preserving consistency with the rest of the skill set was the higher-leverage choice.

All eleven self-review items now closed.

skills/wp-abilities-verify/references/permission-roundtrip.md
rtio: Landed in 697629e.

Added a cleanup block after each result-printing loop in both eval scripts (`permission-roundtrip.md` main runtime block and `runtime-harness.md` Check 5):

```php
if ( isset( $sub_id ) && ! is_wp_error( $sub_id ) ) {
    require_once ABSPATH . "wp-admin/includes/user.php";
    wp_delete_user( $sub_id );
}
```

`wp_delete_user()` lives in `wp-admin/includes/user.php`, which `wp eval` does not auto-load — hence the `require_once`. The caller needs `delete_users`; in `permission-roundtrip.md` the last context exercised is admin so the cleanup runs directly. In `runtime-harness.md` the last context is subscriber, so the cleanup first switches back to admin via `wp_set_current_user( 1 )` before deleting.

The `isset() && ! is_wp_error()` guard mirrors the create check — nothing to clean up if create failed.

skills/wp-abilities-verify/references/permission-roundtrip.md
```diff
@@ -0,0 +1,183 @@
...
+    }
+    echo $context . "=" . $printable . PHP_EOL;
+}
+
+// Cleanup the test subscriber so repeated harness runs don't accumulate users
+// on shared dev environments. Already running as admin (line above), so the
+// caller has the delete_users capability.
+if ( isset( $sub_id ) && ! is_wp_error( $sub_id ) ) {
+    require_once ABSPATH . "wp-admin/includes/user.php";
+    wp_delete_user( $sub_id );
```
rtio: **Low** — the cleanup added in `697629e` works on single-site (the skill's documented baseline — `wp-env (Docker), WordPress 6.9` defaults to single-site) but is a no-op for the user-row deletion on multisite. `wp_delete_user()` only removes the user from the current site's membership; `wp_users` and `wp_usermeta` rows persist at the network level (verified against `wp-admin/includes/user.php:469` in `wordpress-develop` trunk — the docblock confirms this).

Same pattern in `runtime-harness.md` Check 5 around line 241.

What if we add a multisite branch? Note that `wpmu_delete_user()` lives in `wp-admin/includes/ms.php`, not `user.php` — both files need to be required when guarding:

```php
if ( isset( $sub_id ) && ! is_wp_error( $sub_id ) ) {
    if ( is_multisite() ) {
        require_once ABSPATH . 'wp-admin/includes/ms.php';
        wpmu_delete_user( $sub_id );  // network-wide delete
    } else {
        require_once ABSPATH . 'wp-admin/includes/user.php';
        wp_delete_user( $sub_id );    // single-site only
    }
}
```

Or, if multisite isn't in scope for this skill, a one-line note (*"single-site only; multisite requires `wpmu_delete_user`"*) next to the cleanup block keeps the snippet small and warns anyone running it on multisite.

rtio (commented):
## Incremental review — verdict: COMMENT

Follow-up to [`pullrequestreview-4266346998`](https://github.com/WordPress/agent-skills/pull/47#pullrequestreview-4266346998). 10 of 11 prior findings landed cleanly across the 10 new commits — the verify-ignore boundary prose, schema-lints heading reconciliation, `check_permissions( $input )` fix, cross-skill path correction, symbol-name citations, blind-spot table additions, and the bonus Jetpack env-up cleanup all check out across clarity, drift, and WP-architecture lenses.

One residual partial-fix on the test-user cleanup commit — see the inline comment below. Single-site dev environments (the skill's documented baseline) are unaffected; the gap only matters on multisite.

### Reviewers this pass

code-clarity (APPROVE), docs-drift (APPROVE), wp-architecture (1 finding). Decision-critic returned REVISE — caught a bug in the originally drafted recommendation snippet (`wpmu_delete_user` lives in `wp-admin/includes/ms.php`, not `user.php` — the snippet would have fataled on the very environment it was supposed to fix). Both corrected before posting.

### Mapping of prior findings → fix commits

| Prior finding | Commit | Status |
|---|---|---|
| M1 `check_permissions()` no-arg | `2b86824` | Clean — `array()` is correct shape per WP REST schema layer |
| M2 four/three gotchas heading | `57d3d6d` | Clean — split named explicitly |
| M3 `idempotent=false` example row | `f3c4e27` | Clean — anchored to new boundary prose |
| M4 `readonly: true` definition | `f3c4e27` | Clean — same one-sentence anchor |
| M5 typed/untyped helper signature | `1ba84da` | Clean — harmonized to untyped |
| M6 missing `../` cross-skill path | `613fce7` | Clean — resolves correctly |
| M7 hooks-as-writes blind spots | `6b09b72` | Clean — implicit-hook + AS rows added, all symbols verified |
| L1 `: void` style | `b054288` | Clean |
| L2 `verify_sub_*` cleanup | `697629e` | **Partial** — see inline below |
| L3 eval scenario Step 6 wording | `f3c4e27` | Clean — anchored to boundary prose |
| L4 line-number citations | `0b0a0af` | Clean — switched to symbol names entirely (stronger fix) |

The highest-leverage `verify-ignore` boundary edit (`f3c4e27`) is the most visible of the bunch and resolves three downstream sites in one commit — exactly the shape suggested in the prior review.

skills/wp-abilities-verify/references/permission-roundtrip.md
rtio: Fixed in 6e277af.

Both eval blocks now branch on `is_multisite()`:

```php
if ( isset( $sub_id ) && ! is_wp_error( $sub_id ) ) {
    if ( is_multisite() ) {
        require_once ABSPATH . "wp-admin/includes/ms.php";
        wpmu_delete_user( $sub_id );
    } else {
        require_once ABSPATH . "wp-admin/includes/user.php";
        wp_delete_user( $sub_id );
    }
}
```

Both `require_once` paths because the functions live in different admin-includes files (`user.php` vs `ms.php`) — wp-cli's `wp eval` doesn't auto-load either. Comment above the block expanded to call out the network-vs-site divergence in-place, so a reader of the snippet doesn't have to grep core to know why both files are referenced.

Confirmed against `wp_delete_user()`'s docblock in `wordpress-develop` trunk's `wp-admin/includes/user.php`: on multisite it only removes the user from the current site's membership; `wpmu_delete_user()` is the network-wide delete. The single-site call wasn't a no-op for single-site installs (the documented runtime baseline), but the silent no-op on multisite was exactly the footgun shape the harness shouldn't ship with.

rtio (commented):
Looks good — closing out the review thread on this PR.

`6e277af` cleanly resolves the multisite gap from [`pullrequestreview-4267080988`](https://github.com/WordPress/agent-skills/pull/47#pullrequestreview-4267080988). The `is_multisite()` branch is symmetric across both files, the `wpmu_delete_user()` / `wp_delete_user()` pairing is correct, the `require_once` paths point to the right admin-includes files (`ms.php` vs `user.php`), and the in-place comment explains the divergence so a reader doesn't have to grep core.

Verified:
- `wpmu_delete_user()` performs full network-wide deletion — `wp_users` row + `wp_usermeta` + per-site memberships.
- `is_multisite()` is a pure constant check, safe to call anywhere.
- Caller is user 1 (network super-admin by convention) so capability is sufficient on multisite.
- One precision nuance noted but irrelevant in context: `wpmu_delete_user()` has a super-admin protection no-op at `wp-admin/includes/ms.php:159-163`, but the throwaway `verify_sub_<timestamp>` subscriber will never trigger it.

Review cycle on this PR: original (11 findings) → first incremental (10/11 clean, 1 partial) → second incremental (last partial fully resolved). Nothing outstanding from my side. Calling this an approval — not flipping the formal status, just closing the loop in writing.

skills/wp-abilities-verify/references/runtime-harness.md
```diff
@@ -0,0 +1,380 @@
...
+Cross-reference each annotation against the audit's declared value (if an
+audit was provided) AND against the static inventory's declared value.
+Mismatch on either axis → FAIL.
+
+This check complements the static adversarial check from
+`annotation-correctness.md`: static checks the callback's actual
+behavior; runtime checks what the registration hook resolved to at boot
+time. Both must agree for the annotations to be trustworthy.
+
+## Check 3 — each read ability's `execute()` returns OK or a standard-vocabulary WP_Error
```
nerrad: Could this check distinguish smoke execution from high-confidence verification?

While using these skills to implement abilities across several WooCommerce extensions, most of the useful failures came from executing abilities against representative seeded data, not just confirming that `execute()` returned `OK` or a known `WP_Error`. Examples were wrong IDs, cached sentinel values, permission mismatches, filtered labels, and missing capabilities.

Maybe this section can say: synthetic inputs are acceptable for a smoke check, but the preferred/high-confidence runtime path is to seed realistic plugin data, call `wp_get_ability()->execute()` with representative inputs, and assert the expected output shape/privacy behavior.

rtio:
@nerrad — re-review when convenient. Round 3 (the lighter-framing rewrite per your "ship lighter; iterate from real usage" hint) has been up since 2026-05-08. Further slim can be a follow-up rather than a pre-merge gate if that works.

skills/wp-abilities-verify/references/runtime-harness.md
rtio: @nerrad — done in `b6e1bb4`. Check 3 is now split into two named levels:

**Level 1 — smoke execution (synthetic inputs).** Keeps the previous content verbatim. Confirms each read returns `OK` or a vocabulary `WP_Error`. Catches PHP fatals, un-bootstrapped services, registration failures. The notes already there now read as the positive signal *for Level 1*.

**Level 2 — high-confidence verification (representative seeded data).** For each ability whose audit entry declares a non-null `seed_data_needs` (the field added to the schema in PR #46 R2 — `34cfb58`), seed the environment per that field, call the ability with real ids / slugs (not synthetic placeholders), and assert two things: the documented output shape (keys present, types correct) AND the privacy contract (forbidden fields like `full_pan`, `card_number`, `bank_account_number` must not appear). When `seed_data_needs: null`, the harness reports `LEVEL 2: pending (ask the implementer)` and proceeds — dovetails with the WARN-not-FAIL posture for the audit-schema implementation-readiness fields.

Lists the six failure modes Level 2 catches that Level 1 misses (wrong IDs returning the wrong record, cached sentinels, permission mismatches on real vs synthetic IDs, filtered-vs-unfiltered label drift, capabilities surfacing only against real records, curated-output drift). The Level 2 block is also explicitly framed as the one-off harness run that produces the PR-time artifact; the durable home for these per-ability assertions on every CI run is the WooCommerce-extension overlay's `test-the-public-contract.md` (payments-ai-tools PR #21) which codifies the same six-dimension discipline at unit-test time.

nerrad (approved):

jonathanbossenger (approved):

PR #59, opened 2026-06-01
[wp-block-themes]: document theme.json slug-normaliser bug class

mikeyarce:
## Summary

Documents a silent failure in theme.json: WordPress hyphenates preset/custom slugs before emitting CSS vars, so a handwritten `var()` reference to the un-normalised slug resolves to nothing and quietly falls back. No error, layout looks roughly right — invisible without computed-style inspection.

Examples:

- slug `3xl` → `--wp--preset--font-size--3-xl`
- slug `cardShadow` → `--wp--custom--card-shadow`

Under the hood, `WP_Theme_JSON` runs every slug through `_wp_to_kebab_case()`, which splits at digit/letter boundaries, camelCase transitions, and non-alphanumeric characters.

## Changes

- **`skills/wp-block-themes/references/theme-json.md`** — new "Slug normalisation gotcha" section with the explanation and a grep pattern for catching un-normalised references.
- **`skills/wp-block-themes/references/debugging.md`** — new bullet under "Styles not applying" pointing at the gotcha section.


karmatosed (approved):
Looks good to merge from me and in checking.
We do need 2 approvals though to commit, so let's get another.

bgrgicak (approved):

PR #98, opened 2026-08-24
Fix skill catalog drift in CI, router, and eval coverage

### LINKED ISSUE #97: Skill catalog has drifted: several skills are missing from CI validation, router, and eval coverage
## Summary

Three separate gaps, all with the same root cause — new skills got added to `skills/` without updating the other places that track the full skill list:

### A. CI's `skills-ref validate` step only covers 10 of 18 skills

`.github/workflows/ci.yml` hardcodes 10 skill names. Missing: `blueprint`, `wp-abilities-audit`, `wp-abilities-verify`, `wp-patterns`, `wp-phpstan`, `wp-plugin-directory-guidelines`, `wp-rest-api`, `wpds`. Nearly half the skill set ships with **no external schema validation in CI**.

### B. Router decision tree doesn't route to 5 existing skills

`skills/wordpress-router/references/decision-tree.md` has no entry for `wpds`, `wp-abilities-audit`, `wp-abilities-verify`, `wp-plugin-directory-guidelines`, or `wp-patterns` — all real, shipped skills listed in the README. It also routes "security" intent to a `wp-security (planned)` skill that doesn't exist, even though `wp-plugin-development` already covers that ground.

### C. `wpds` has zero eval scenarios

No file in `eval/scenarios/*.json` mentions `wpds`, violating the authoring guide's own golden rule ("Don't add a new skill without at least one scenario in `eval/scenarios/`").

NoumaanAhamed:
Title:
## Summary
- `.github/workflows/ci.yml` now loops over every `skills/*/` directory for `skills-ref validate` instead of a hardcoded 10-skill list — 8 skills (`blueprint`, `wp-abilities-audit`, `wp-abilities-verify`, `wp-patterns`, `wp-phpstan`, `wp-plugin-directory-guidelines`, `wp-rest-api`, `wpds`) were shipping with zero CI schema validation
- `skills/wordpress-router/references/decision-tree.md` gains routes for `wpds`, `wp-abilities-audit`, `wp-abilities-verify`, `wp-patterns`, and `wp-plugin-directory-guidelines` — previously dead ends with no route
- Security-intent routing now points to `wp-plugin-development` (which already covers nonces/capabilities/sanitization) instead of the nonexistent `wp-security`
- Added `eval/scenarios/wpds-build-component-with-tokens.json` — `wpds` had zero eval coverage

## Test plan
- [x] `node eval/harness/run.mjs` passes
- [x] New CI loop enumerates all 18 skill directories (verified locally via `comm` diff against `ls skills`)
- [x] `eval/scenarios/wpds-build-component-with-tokens.json` validated as well-formed JSON matching the scenario schema

Closes #97

### DIFF
--- a/docs/skill-set-v1.md
+++ b/docs/skill-set-v1.md
@@ -10,12 +10,19 @@ This repo currently includes:
 - `wp-rest-api`
 - `wp-interactivity-api`
 - `wp-abilities-api`
+- `wp-abilities-audit`
+- `wp-abilities-verify`
+- `wp-patterns`
 - `wp-wpcli-and-ops`
 - `wp-performance`
 - `wp-phpstan`
+- `wp-playground`
+- `blueprint`
+- `wpds`
+- `wp-plugin-directory-guidelines`
 
 Planned next skills (not yet implemented):
 
 - `wp-build-tooling`
 - `wp-testing`
-- `wp-security`
+- `wp-security` (a dedicated split-out of the security guidance currently covered by `wp-plugin-development`)
--- a/skills/wordpress-router/references/decision-tree.md
+++ b/skills/wordpress-router/references/decision-tree.md
@@ -25,6 +25,16 @@ Route by intent even if repo kind is broad (like `wp-site`):
   - Route → `wp-interactivity-api`.
 - **Abilities API / wp_register_ability / wp-abilities/v1 / @wordpress/abilities**
   - Route → `wp-abilities-api`.
+- **Audit a plugin's REST surface and propose Abilities API registrations**
+  - Route → `wp-abilities-audit`.
+- **Verify a plugin's Abilities API registrations against their declared annotations (e.g. readonly-but-writes check)**
+  - Route → `wp-abilities-verify`.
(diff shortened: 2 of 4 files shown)


docs/skill-set-v1.md
justlevine, line 1: The solution here is likely ditching the `-v1` slop from the doc (or ditching this _entire file_ as ai-slop), but if not, then it begs the question what makes these new skills part of a "v1"?

skills/wordpress-router/references/decision-tree.md
```diff
@@ -52,7 +62,7 @@ Route by intent even if repo kind is broad (like `wp-site`):
 - **Performance / caching / query profiling / editor slowness**
   - Route → `wp-performance`.
 - **Security / nonces / capabilities / sanitization/escaping / uploads**
-  - Route → `wp-security` (planned).
+  - Route → `wp-plugin-development` (already covers this today); a dedicated `wp-security` split-out is planned but not yet implemented.
```
justlevine, line 65: This is not an AI-decision log. Either it should route to somewhere specific or it shouldn't because it's "planned". but if it's "already covered", then there is 0 human reason to include that it's covered (self-documenting) or potential future refactor plans.

